PR #12, opened 2021-01-11
feat(snapshot): add snapshot support for LVM PV

### LINKED ISSUE #10: add snapshot support for lvm-localpv

We want to be able to create a snapshot of the lvm volumes.

akhilerm:
This PR adds support for LVM snapshot to the lvm-localPV CSI driver. The snapshots created will be readonly (as opposed to the default ReadWrite). Also, once snapshots are  created for a volume, resize will not work for those volumes, since LVM does't support that.

To create a snapshot, create a snapshot class as given below and then create a `volumesnapshot` resource
```
kind: VolumeSnapshotClass
apiVersion: snapshot.storage.k8s.io/v1
metadata:
  name: lvm-localpv-snapclass
  annotations:
    snapshot.storage.kubernetes.io/is-default-class: "true"
driver: local.csi.openebs.io
deletionPolicy: Delete
---
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: lvm-localpv-snap
spec:
  volumeSnapshotClassName: lvm-localpv-snapclass
  source:
    persistentVolumeClaimName: <pvc-name>
```

**What this PR does?**:
- adds LVMSnapshot CRDs
- add snapshot controller to watch for LVMSnapshot CRs
- adds the volumesnapshot related CRDs from storage.k8s.io to the deployment
- use container images from `k8s.gcr.io` for CSI components

**Limitation**
Volumes with snapshots cannot be resized, as LVM does not support online resize of origin volumes with a snapshot. `ControllerExpandVolume` will error out if the volume to resized has any active snapshots.

`dm-snapshot` kernel module should be loaded for snapshot to work

1. Snapshot creation
2. Try to resize volume with snapshot (will error out the volume expansion)
3. resize should work after snapshots are removed
4. create multiple snapshots for the same volume 

**Checklist:**
- [x] Fixes #10 
- [x] Has the change log section been updated?
- [ ] Commit has unit tests


deploy/yamls/lvm-driver.yaml
```diff
@@ -28,67 +28,65 @@ spec:
...
+            - deletionPolicy
+            - driver
           type: object
-      required:
-      - deletionPolicy
-      - driver
-      type: object
-  version: v1beta1
-  versions:
-  - name: v1beta1
```
akhilerm, line 76: @pawanpraka1 @prateekpandey14 . This is the part that would be removed, if we remove v1beta1.

akhilerm: This means the CRD definition itself in the system will not have v1beta1 present.

pkg/driver/agent.go
```diff
@@ -58,6 +59,14 @@ func NewNode(d *CSIDriver) csi.NodeServer {
 		}
 	}()
 
+	// start the lvm snapshot watcher
+	go func() {
+		err := snapshot.Start(&ControllerMutex, stopCh)
+		if err != nil {
+			klog.Fatalf("Failed to start ZFS volume snapshot management controller: %s", err.Error())
```
kmova, line 66: n_: `ZFS` to LVM.

pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
```diff
@@ -65,6 +65,8 @@ type VolumeInfo struct {
 	// +kubebuilder:validation:MinLength=1
 	VolGroup string `json:"volGroup"`
 
+	SnapName string `json:"snapName,omitempty"`
```
pawanpraka1, line 68: why do we need this?

pkg/mgmt/snapshot/snapshot.go
```diff
@@ -0,0 +1,247 @@
...
+	// LVMSnapshot should be deleted. Check if deletion timestamp is set
+	if c.isDeletionCandidate(snap) {
+		err = lvm.DestroySnapshot(snap)
+		if err == nil {
+			lvm.RemoveSnapFinalizer(snap)
+		}
+	} else {
+		// if finalizer is not set then it means we are creating
+		// the lvm snapshot.
+		if snap.Finalizers == nil {
```
pawanpraka1, line 87: can we rely on snap.Spec.Status to be Pending here?

pkg/driver/controller.go
```diff
@@ -444,7 +445,71 @@ func (cs *controller) CreateSnapshot(
 	req *csi.CreateSnapshotRequest,
 ) (*csi.CreateSnapshotResponse, error) {
 
-	return nil, status.Error(codes.Unimplemented, "")
+	klog.Infof("CreateSnapshot volume %s for %s", req.Name, req.SourceVolumeId)
+
+	snapTimeStamp := time.Now().Unix()
```
pawanpraka1, line 450: can we add verifySnapshotRequest to verify all the csi compliant things there (see the latest zfs-localpv code https://github.com/openebs/zfs-localpv/blob/master/pkg/driver/controller.go#L581)

pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
akhilerm: This was added for clone support, same as from ZFS LocalPV

akhilerm: removed.

pkg/driver/controller.go
akhilerm: Done.

akhilerm:
@pawanpraka1 Updated the `NodeExpandVolume` to error out if the volume contains any snapshots

pawanpraka1:
@akhilerm we need to block that in ControllerExpandVolume.

akhilerm:
> @akhilerm we need to block that in ControllerExpandVolume. Also test this and see if behavior is fine.

Done. @pawanpraka1. Yes , tested and its working fine.

One issue I faced is in the following test case:
1. Create PVC of 4Gi
2. Create a snapshot
3. Resize PVC to 5Gi
4. Resize fails because of active snapshot
5. Delete the snapshot
6. But the controller logs is not having the ControllerExpandVolume call again.
7. Even after deleting the snapshot the resize is not happening.

pawanpraka1:
@akhilerm it uses exponential backoff to retry the next time. You might need to wait enough to get the call again?

akhilerm:
@pawanpraka1 Changing the grpc error code worked.

pkg/mgmt/snapshot/snapshot.go
akhilerm: @pawanpraka1  Which is the case that we are hitting here?

pawanpraka1: rather than checking whether finalizer us set it is better to check if request is pending then go ahead and create the snapshot.

pkg/driver/controller.go
```diff
@@ -635,3 +747,38 @@ func (cs *controller) validateVolumeCreateReq(req *csi.CreateVolumeRequest) erro
 	}
 	return nil
 }
+
+func (cs *controller) validateSnapshotRequest(req *csi.CreateSnapshotRequest) error {
```
pawanpraka1, line 751: I don't see thais getting called in the CreateSnapshot request. Where we are validating snapshot request?

akhilerm: Updated.

pkg/mgmt/snapshot/snapshot.go
akhilerm: Yep. we can do that. Why is this approach not followed in zfs localPV?

akhilerm: Got the answer : "as there was no status initially, finalizer was used to check if the snapshot was created."

Updated the changes.

deploy/yamls/lvm-driver.yaml
```diff
@@ -548,8 +758,26 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
+        - name: csi-snapshotter
+          image: quay.io/k8scsi/csi-snapshotter:v4.0.0
```
pawanpraka1, line 762: can we change in image to gcr?

deploy/yamls/lvm-driver.yaml
```diff
@@ -548,8 +758,26 @@ spec:
...
+            - "--csi-address=$(ADDRESS)"
+            - "--leader-election"
+          env:
+            - name: ADDRESS
+              value: /var/lib/csi/sockets/pluginproxy/csi.sock
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /var/lib/csi/sockets/pluginproxy/
+        - name: snapshot-controller
+          image: quay.io/k8scsi/snapshot-controller:v4.0.0
```
pawanpraka1, line 774: here also?

deploy/yamls/lvm-driver.yaml
```diff
@@ -720,7 +948,7 @@ spec:
       hostNetwork: true
       containers:
         - name: csi-node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:v1.2.0
+          image: quay.io/k8scsi/csi-node-driver-registrar:v2.1.0
```
pawanpraka1, line 951: gcr image.

akhilerm: Done.

pawanpraka1 (approved):
looks good.

PR #14, opened 2021-01-25
feat(block): adding raw block volume support for LVM

pawanpraka1:

adding raw block volume support for LVM

No

tested with the sample fio yaml added in this PR

### DIFF
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -113,7 +113,7 @@ func (ns *node) NodePublishVolume(
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	// If the access type is block, do nothing for stage
+
 	switch req.GetVolumeCapability().GetAccessType().(type) {
 	case *csi.VolumeCapability_Block:
 		// attempt block mount operation on the requested path
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -17,6 +17,7 @@ limitations under the License.
 package driver
 
 import (
+	"os"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


pkg/driver/agent.go
```diff
@@ -115,7 +115,8 @@ func (ns *node) NodePublishVolume(
 	// If the access type is block, do nothing for stage
```
akhilerm, line 115: Didnt get this comment. This is the publish request.

pawanpraka1: this is not a vaild comment here, this is for nodestagevolume grpc call.

pkg/lvm/mount.go
```diff
@@ -199,3 +199,32 @@ func MountFilesystem(vol *apis.LVMVolume, mount *MountInfo) error {
...
 	return MountVolume(vol, mount)
 }
+
+// MountBlock mounts the block disk to the specified path
+func MountBlock(vol *apis.LVMVolume, mountinfo *MountInfo) error {
+	target := mountinfo.MountPath
+	volume := vol.Spec.VolGroup + "/" + vol.Name
+	devicePath := DevPath + volume
+
+	mountopt := []string{"bind"}
```
akhilerm, line 209: Why do we need to bind mount the device at a path during nodepublish. Is it a requirement from CSI?

akhilerm (commented):
given a comment regarding workflow.

pkg/lvm/mount.go
pawanpraka1: yes, we need to moount the device to the path where kubelet wants us to.

pkg/driver/agent.go
akhilerm: @pawanpraka1  Can you remove this comment,

akhilerm (approved):
LGTM. Gave a suggestion on removing the comment.

kmova (approved):

PR #21, opened 2021-02-11
feat(scheduling): add volume group capacity tracking

iyashu:
With k8s version >= 1.19, there is feature(alpha stage) added where the kube-scheduler gets influenced by storage capacity available on nodes during filtering stage. Without [CSIStorageCapacity](https://kubernetes.io/docs/concepts/storage/storage-capacity/) tracking, a pod with a volume that uses delayed binding may get scheduled multiple times, but might always land on the same node unless there are multiple nodes with equal priority. More details are mentioned in [KEP 1472](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1472-storage-capacity-tracking) & [external provisioner](https://github.com/kubernetes-csi/external-provisioner#capacity-support) documentation.

**What this PR does?**:
We introduced a new custom resource called `LVMNode` recording all the available vgs and its corresponding attributes in each node. Openebs node plugin running on each node will periodically scan vgs and reconcile `LVMNode` resource. At controller side, we've implemented CSI GetCapacity method called by k8s external-provisioner for creating/updating CSIStorageCapacity objects which in turn used by kube-scheduler for each node topology. K8s external-provisioner calls GetCapacity for every combination of storage class and node topology segments (individual nodes in our case). So, volume group configured in storage class is also a part of args in GetCapacity method. In case, there are multiple volume group (with multi pool support) available on nodes, we choose the volume group having maximum free capacity & its name matching to volgroup parameter of the GetCapacity call.

No

Consider a kubernetes cluster having 5 worker nodes (besides master ones) each having a storage capacity of size 32 Gi. Create a stateful set cluster say `sts-a` of size 4 each requesting storage capacity(pvc) of size 20Gi. Now create another stateful set cluster say `sts-b` of size 1 again requesting storage capacity(pvc) of size 20Gi. We'll see that pod in `sts-b` will be scheduled (by kube-scheduler) on right node having enough capacity.

**Checklist:**
- [x] Fixes #<issue number>
- [ ] Commit has unit tests


iyashu:
Gentle reminder @pawanpraka1 @akhilerm

pkg/apis/openebs.io/lvm/v1alpha1/lvmnode.go
```diff
@@ -0,0 +1,80 @@
...
+// in a node. In general, the openebs node-agent creates the LVMNode
+// object & periodically synchronizing the volume groups available in the node.
+// LVMNode has an owner reference pointing to the corresponding node object.
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=lvmnode
+type LVMNode struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	VolumeGroups []VolumeGroup `json:"volumeGroups"`
```
akhilerm, line 38: Shouldn't the `VolumeGroups` be in the `Spec` section as per the convention in k8s API resources.

pkg/driver/agent.go
```diff
@@ -52,6 +53,13 @@ func NewNode(d *CSIDriver) csi.NodeServer {
 	// set up signals so we handle the first shutdown signal gracefully
 	stopCh := signals.SetupSignalHandler()
```
akhilerm, line 55:
```suggestion
// start the lvm node resource watcher
```

pkg/driver/controller.go
```diff
@@ -18,16 +18,27 @@ package driver
 
 import (
 	"fmt"
-	"github.com/openebs/lvm-localpv/pkg/builder/snapbuilder"
```
akhilerm, line 21: :+1:

pkg/apis/openebs.io/lvm/v1alpha1/lvmnode.go
iyashu: In general yes, but not always. `Spec` is used to record the intent of resource & `Status` specify the current state of world (as seen by controllers). Here the `LVMNode` crd is informative rather than any spec (user intend per se) where controller tries to match it with external state. See [api-conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status) for more details. There are many crds which doesn't follow spec/status pattern unless required. For example, in this context see [CSIStorageCapacity crd](https://github.com/kubernetes/kubernetes/blob/6d5cb36d36f34cb4f5735b6adcd5ea8ebb4440ba/staging/src/k8s.io/api/storage/v1alpha1/types.go#L160) which is again informative rather than a user intend.

pkg/lvm/lvm_util.go
```diff
@@ -265,3 +271,82 @@ func DestroySnapshot(snap *apis.LVMSnapshot) error {
...
+	}
+	return vg, nil
+}
+
+// ListLVMVolumeGroup invokes `vgs` to list all the available volume
+// groups in the node.
+func ListLVMVolumeGroup() ([]apis.VolumeGroup, error) {
+	args := []string{
+		"--options", "vg_all",
+		"--reportformat", "json",
```
akhilerm, line 342: Need to add a minimum system requirement, because in older versions of Ubuntu 16.04 , `reportformat` option is not supported.

pkg/lvm/lvm_util.go
```diff
@@ -265,3 +271,82 @@ func DestroySnapshot(snap *apis.LVMSnapshot) error {
 func getLVMSnapName(snapName string) string {
 	return strings.TrimPrefix(snapName, "snapshot-")
 }
+
+func decodeVgsJSON(raw []byte) ([]apis.VolumeGroup, error) {
+	output := &struct {
+		Report []struct {
+			VolumeGroups []map[string]string `json:"vg"`
```
akhilerm, line 278: If we specify all the json keys correctly, cant we directly unmarshal to `apis.VolumeGroup` ? or some other struct instead of using `[]map[string]string`. It is confusing.

iyashu: Issue is lvm json output always returns json objects values in string regardless if it's numeric(e.g vg size) or boolean which is unfortunate. See the example below. So, it may not a good idea to mask the typing(int/bool/string) or size SI units information with fields under `apis.VolumeGroup` just for the benefit of easy parsing since in that case one will end up parsing these string fields everytime as & when we'll be using/consuming it (e.g in GetCapacity we need to compare vg size).

```
{
  "report": [
    {
      "vg": [
        {
          "vg_fmt": "lvm2",
          "vg_uuid": "VPPIu3-FCFI-svbR-lljg-aqID-ryuN-24ShtQ",
          "vg_name": "lvmvg",
          "vg_attr": "wz--n-",
          "vg_permissions": "writeable",
          "vg_extendable": "extendable",
          "vg_size": "32.00g",
          "vg_free": "32.00g",
          "vg_extent_size": "4.00m",
          "vg_extent_count": "8191",
          "vg_free_count": "8191",
          "pv_count": "1",
          "lv_count": "0",
          "snap_count": "0",
          "vg_seqno": "23"
        }
      ]
    }
  ]
}
```

pkg/lvm/lvm_util.go
iyashu: Thanks for pointing out. In that case, do we need to support 16.04? It's getting EOLed in april 21. We can take a call of supporting only lvm versions greater than certain value. As per the [RHEL documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/logical_volume_manager_administration/custom_report#json-output), it's being supported since redhat v7.3. I need to check how does it gets translated to lvm driver version. If reportformat flag is not supported, I presume that command will fail and user will not be able to use this capacity tracking feature. Parsing the default output format (tabular) will result in more tedious exercise & i'm not sure if it's worth the effort since 16.04 will be eoled in two months.

pkg/apis/openebs.io/lvm/v1alpha1/register.go
```diff
@@ -73,6 +73,9 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&LVMVolumeList{},
 		&LVMSnapshot{},
 		&LVMSnapshotList{},
+
```
prateekpandey14, line 76: nit: remove extra line

pkg/lvm/lvm_util.go
akhilerm: @pawanpraka1 Can we decide on the minimum supported version

deploy/lvm-operator.yaml
```diff
@@ -246,6 +246,118 @@ status:
...
+###########     LVMNode CRD       ############
+###########                       ############
+##############################################
+
+# LVMNode CRD is autogenerated via `make manifests` command.
+# Do the modification in the code and run the `make manifests` command
+# to generate the CRD definition
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
```
prateekpandey14, line 261: we can start using `v1` CRDs version ..`v1beta` going to be deprecated in k8s 1.22

deploy/lvm-operator.yaml
```diff
@@ -246,6 +246,118 @@ status:
...
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.8
```
prateekpandey14, line 265: We can start using the controller-gen verison v0.4.0 version which will basically generates the `v1` CRD version by default.

pkg/mgmt/lvmnode/builder.go
```diff
@@ -0,0 +1,157 @@
+/*
+Copyright 2019 The OpenEBS Authors
```
prateekpandey14, line 2: nit: 2021 ?

deploy/lvm-operator.yaml
iyashu: Yeah, I had initially configured the same, then saw that other crds (LVMSnapshot, LVMVolume etc) are using v1beta1. I think, we can take this up as part of separate pull request as it'll required changes to other crds as well.

iyashu: Yeah, I had initially configured the same, then saw that other crds (LVMSnapshot, LVMVolume etc) are using v1beta1. I think, we can take this up as part of separate pull request as it'll required changes to other crds as well.

pkg/lvm/lvm_util.go
akhilerm: This wont be an issue as we are using the LVM binary from the alpine image. So the version will not be a problem.

pkg/apis/openebs.io/lvm/v1alpha1/lvmnode.go
prateekpandey14: Are we not going to have the `Status` for  `lvmNode` resource ? IMO  then  it make sense to not have `Spec` Section

pkg/mgmt/lvmnode/lvmnode.go
```diff
@@ -0,0 +1,307 @@
...
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("invalid resource key: %s", key))
+		return nil
+	}
+
+	return c.syncNode(namespace, name)
+}
+
+// synNode is the function which tries to converge to a desired state for the
```
prateekpandey14, line 53: nit: syncNode

akhilerm (approved):
LGTM

prateekpandey14 (approved):
LGTM, some of the changes like CRDs version etc going to be fixed in upcoming PRs

pkg/driver/agent.go
```diff
@@ -53,6 +54,14 @@ func NewNode(d *CSIDriver) csi.NodeServer {
...
 
+	// start the lvm node resource watcher
```
pawanpraka1, line 57: why do we need to have controller for lvm node, there is no need of having watcher for the object.

pawanpraka1 (commented):
@iyashu why do we need watcher for lvmnode object? does node daemonset need to take any action as per modification/updation of lvmnode object?

pkg/driver/agent.go
iyashu: Yes, it's possible to keep lvm node resource in sync without the controller. But there are few things I had in mind at that time before taking up the controller path. Informers are anyways required for upserting the lvmnode objects. In absence  of informers, we'll end up querying api server (thus increasing load on control plane) for every periodic update since we need to first fetch and then update the resource if required. It'll becomes more visible in case of aggressive sync period say 5s in large clusters. Only extension I can think of between controllers and informers is usage of worker queue to process each event on resource (including periodic resync). Now, currently we reconcile lvm node object almost immediately in case admin updates/deletes that resource by mistake rather than waiting for next sync period. (As I can now, the [CSIStorageCapacity](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/capacity.go#L271) also uses similar controller based approach to reconcile immediately)

Since all the other CRDs (lvm volume & lvm snapshot) are building using the similar pattern, I thought of using the same   constructs for better understandability. But I agree, we could have used just informers cache and write a loop which updates these lvmnode resources periodically. I'm wondering do we see any specific issue with this controller approach?

pawanpraka1: > we could have used just informers cache and write a loop which updates these lvmnode resources periodically

yes, it makes sense to use informer cache and a write loop. watcher does not make any sense here as node daemonset does not have to do anything in response to update/delete of the lvmnode resource. Can we take this as enhancement in the next release, I will go ahead and merge this PR?

iyashu: Sure. We can go ahead with current implementation & discuss the enhancement in next meeting. For curiosity purpose, I see the [CSIStorageCapacity](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/capacity/capacity.go#L271) (a very similar use case as ours) resources managed by external provisioner also reconcile immediately in case of any rouge update or deletion of these objects by say cluster admin. I think, with reconciling immediately we are just reducing the blast radius of stale information (although staleness is possible with other scenarios). Anyways, let's discuss this in next meeting.

PR #23, opened 2021-02-17
feat(provisioning) enable pod resheduling cause of node insufficient capacity

### LINKED ISSUE #18: Add ability to reschedule the volume provisioning in case selected node doesn't have enough capacity
[A description of the current limitation/problem/challenge that you are experiencing.]
As of now, if the [selected node](https://github.com/openebs/lvm-localpv/blob/b370f3c3ea8f69467a23e8e90139e81b9e2ce632/pkg/driver/controller.go#L234) vg doesn't have enough capacity, pod will remain in pending (when using wait param) or container creating state forever. Instead of that, one needs to propagate ResourceExhausted error code to csi-provisioner, so that it can possibly reschedule the pod to some other node. See [this](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume-errors) & [this](https://github.com/kubernetes-csi/external-provisioner/blob/637608cab7d37ddef0575fa2191ba22c57e43cc2/pkg/controller/controller.go#L1671) for more details.

[A clear and concise description of what you want to happen.]
Inspect the available capacity before provisioning the logical volume at node agent and if node vg doesn't have enough capacity, then propagate the ResourceExhausted error to controller (probably by introducing some error code field in lvmvol crd). In case of `wait` params set, the controller return the same error code to caller (i.e csi-external-provisioner).

**Environment:**
- LVM Driver version

iyashu:
As of now, if the volume provisioning fails due to insufficient capacity on nodes, the pod will remain in pending or container creating state forever (unless someone recreates it). See #18 for more details. Apart from this, the pull request also gracefully handles the rpc timeouts & internal errors ensuring idempotency in provisioning workflow.

**What this PR does?**:
We introduce additional field named as `VolumeError` in `LVMVolume` custom resource to propagate the error from node agent to controller. Additionally, node agent will mark the lvm volume being failed if no sufficient space is available in node vg. Controller will inspect the errors populated by node agent and according return either `ResourceExhausted` error code, so that [external provisioner](https://github.com/kubernetes-csi/external-provisioner/blob/f937d0d4a1d88d866d95fad51148b686d9357cfc/pkg/controller/controller.go#L1671) can reschedule the pod. In other cases, controller returns status `Aborted` to treat the volume provisioning in progress.

**Note:** Pull request currently doesn't handles rescheduling in case of implicit binding. It should be trivial & I'll create a separate pr for the same.

No

Consider a cluster where only one node has sufficient capacity to fulfill the pvc capacity request. Now, if you try to create a pod (with late binding), kube-scheduler will keep retrying scheduling if node doesn't have enough capacity available in volume group until it finds the right node.

**Checklist:**
- [x] Fixes #18 
- [ ] Commit has unit tests


pkg/driver/params.go
```diff
@@ -0,0 +1,95 @@
...
+		value, ok := m[key]
+		if !ok {
+			continue
+		}
+		*param = value
+	}
+
+	// parse bool params
+	boolParams := map[string]*bool{
+		"wait": &params.WaitForProvision,
```
pawanpraka1, line 75: do we need this parameter? we will always wait for the provisioning.

pkg/lvm/lvm_util.go
```diff
@@ -89,6 +112,7 @@ func CreateVolume(vol *apis.LVMVolume) error {
 	out, err := cmd.CombinedOutput()
 
 	if err != nil {
+		err = newExecError(out, err)
```
pawanpraka1, line 115: should we use newExecError in the previous return at line 103?

pkg/driver/controller.go
```diff
@@ -101,109 +101,127 @@ func getRoundedCapacity(size int64) int64 {
...
-		switch vol.Status.State {
-		case lvm.LVMStatusReady:
-			return nil
+	// Now it must be in failed state if not above. See if we need
+	// to reschedule the lvm volume.
+	var errMsg string
+	if volErr := vol.Status.Error; volErr != nil {
+		errMsg = volErr.Message
+		if volErr.Code == lvmapi.InsufficientCapacity {
+			reschedule = true
```
pawanpraka1, line 130: can we make is reschedule everytime there is failure on the  node, does not matter for what reason. If it fails on a perticular node, we would want to try the volume creation on some other node.

iyashu: I think it may not be a good idea to reschedule volume in case of any (possibly intermittent) error. It'll result in sparse & unpredictable scheduling of storage resources. IMO, if there is any error related to a particular node, it's better to inspect & fix that error(by cordoning the node) rather than masking it with rescheduling.

pkg/lvm/lvm_util.go
iyashu: It's not required. As the name suggests, newExecError just wraps only the error returned by `exec` pkg after running the cmd. Wrapping other errors will not provide any extra information since output will be empty.

pkg/driver/params.go
iyashu: Yep, removed.

pkg/mgmt/volume/volume.go
```diff
@@ -74,28 +75,58 @@ func (c *VolController) enqueueVol(obj interface{}) {
...
+		return nil
+	}
+
+	if err = lvm.CreateVolume(vol); err != nil {
+		// validate if it's insufficient space error & accordingly set up the error code.
+		if volErr := c.transformLVMError(err); volErr.Code == apis.InsufficientCapacity {
+			vol.Status.Error = volErr
+			err = lvm.UpdateVolInfo(vol, lvm.LVMStatusFailed)
 		}
+		return err
```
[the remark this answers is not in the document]
iyashu: Yes, it is intentional. As of now, we are setting up the failed state only in case of Insufficient capacity where we don't want to retry the volume provisioning. In other cases, controller loop will keep retrying the volume provisioning.

pkg/driver/schd_helper.go
```diff
@@ -17,9 +17,10 @@ limitations under the License.
 package driver
 
 import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"strconv"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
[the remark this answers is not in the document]
iyashu: I'm not sure if we are using gofmt for maintaining a same code format. [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) (being superset of gofmt) separates standard lib imports with third party ones by a newline. More details are [here](https://hyr.mn/gofmt/).

pawanpraka1:
@iyashu there is conflict in the file pkg/driver/controller.go. Cab you rebase and push.

iyashu:
@pawanpraka1 done, ptal.

deploy/lvm-operator.yaml
```diff
@@ -123,14 +123,28 @@ spec:
           description: VolStatus string that specifies the current state of the volume
             provisioning request.
           properties:
+            error:
```
pawanpraka1, line 126: do we need this now? we can directly rely on state being failed to take the decesion?

iyashu: Yeah, functionally from rescheduling point of view it can be removed now. I had just kept it for the purpose of better inspection as user/maintainer. As we are propagating error message to csi external provisioner (along with ResourceExhausted), the same error message is being added as events under pvc resource. e.g 
```
➜  lvm-localpv git:(storage-capacity) ✗ k describe pvc claim-a-test-a-4
Name:          claim-a-test-a-4
Namespace:     default
StorageClass:  openebs-ext4pv
Status:        Pending
Volume:
Labels:        app=test-a
Annotations:   volume.beta.kubernetes.io/storage-provisioner: local.csi.openebs.io
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:
Access Modes:
VolumeMode:    Filesystem
Used By:       test-a-4
Events:
  Type     Reason                Age               From                                                                                Message
  ----     ------                ----              ----                                                                                -------
  Normal   WaitForFirstConsumer  22s               persistentvolume-controller                                                         waiting for first consumer to be created before binding
  Normal   Provisioning          5s (x4 over 22s)  local.csi.openebs.io_openebs-lvm-controller-0_6f7c487b-6397-4d19-a5d4-b4d84819f831  External provisioner is provisioning volume for claim "default/claim-a-test-a-4"
  Normal   ExternalProvisioning  5s (x5 over 22s)  persistentvolume-controller                                                         waiting for a volume to be created, either by external provisioner "local.csi.openebs.io" or manually created by system administrator
  Warning  ProvisioningFailed    3s (x4 over 19s)  local.csi.openebs.io_openebs-lvm-controller-0_6f7c487b-6397-4d19-a5d4-b4d84819f831  failed to provision volume with StorageClass "openebs-ext4pv": rpc error: code = ResourceExhausted desc =   Volume group "lvmvg" has insufficient free space (3071 extents): 5120 required.
 - exit status 5
  Normal  WaitForPodScheduled  3s (x5 over 19s)  persistentvolume-controller  waiting for pod test-a-4 to be scheduled
```

pawanpraka1 (approved):
looks good. Can you describe how you have tested it? Can you have a setup and request for a size which only one node can serve and see if it is getting/provisioner tried on this node eventually?

iyashu:
> looks good. Can you describe how you have tested it? Can you have a setup and request for a size which only one node can serve and see if it is getting tried/provisioner on this node eventually?

Exactly see the pull request description above (If the changes in this PR are manually verified, list down the scenarios covered). In my case, I've a 4 nodes cluster (besides master nodes) each having storage capacity of 32 GiB. I've created 4 pods each requesting storage capacity of size 20 GiB. Now, I can see from lvm node plugin logs that kube-scheduler first tries to schedule all 4 pods on first 3 nodes, but then eventually it schedule the 4th pod on that remaining node (having sufficient capacity).

Another test that I've run is when there is no nodes available to satisfy the pvc claim. In that case, kube-scheduler keeps on retrying the pod across nodes until any nodes gets capacity enough to fit the claim size.

pkg/driver/controller.go
```diff
@@ -101,107 +101,122 @@ func getRoundedCapacity(size int64) int64 {
...
-func waitForVolDestroy(volname string) error {
-	for true {
-		_, err := lvm.GetLVMVolume(volname)
-		if err != nil {
-			if k8serror.IsNotFound(err) {
-				return nil
-			}
-			return status.Errorf(codes.Internal,
-				"lvm: destroy wait failed, not able to get the volume %s %s", volname, err.Error())
+	if reschedule {
```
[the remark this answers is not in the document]
iyashu: We can since reschedule var is only set to true at one place . Just keeping it here for readability purpose where we first take decision on whether we want to reschedule or not and then at last we run some clean up actions based on if rescheduling is required. I can move it if you think otherwise.

pawanpraka1:
@sounds good @iyashu .Thanks for this PR.

PR #25, opened 2021-02-22
fix(gomod): updating lib-csi go mod to v0.2.0

pawanpraka1:
v0.2.0 has new scheduler changes where it returns list of nodes instead of single node. Updated the go mod and the code to pick the first selected node in the list.


shubham14bajpai (approved):
LGTM

akhilerm (approved):
LGTM

PR #27, opened 2021-03-03
feat(action): migrating to github action from travis

pawanpraka1:
removed the .travis file and moved all the bdd test cases to github action.


akhilerm (commented):
@pawanpraka1 A few comments:
1. Should the BDD tests be done for the build workflow also?
2. The travis.yml file can be deleted.

pawanpraka1:
@akhilerm 
> Should the BDD tests be done for the build workflow also?

I thought about it did not see the benefit of doing that.

> The travis.yml file can be deleted.
Have not deleted it purposely as there are now ansible images added from e2e team in the repo, it is not using multiarch build.

akhilerm:
> I thought about it did not see the benefit of doing that.

If all the PRs will be always rebased and merged, then its not needed. But we cannot guarantee that. Consider the case where, change X in PR-1 and change Y in PR-2. The BDD will be tested on both of them separately. Suppose PR-2 is merged and then PR-1 is merged, without rebasing, then the combined changes X and Y wont be tested in the BDD. If we include the bdd in the build workflow, we can guarantee that the changes in the master is tested.

> Have not deleted it purposely as there are now ansible images added from e2e team in the repo, it is not using multiarch build.
Single platform images can also be built on github actions. Just  like how we build normally in our local setup.

pawanpraka1:
> If we include the bdd in the build workflow, we can guarantee that the changes in the master is tested.

will add for build also.

> Single platform images can also be built on github actions. Just like how we build normally in our local setup.

We need to have DNMAE and DPASS setting in the github action, or another way could be if we change the code to use the DOCKER-HUB token and password.

pawanpraka1:
Also what about quay and the images which are built with arch name?

akhilerm:
> Also what about quay and the images which are built with arch name?

Images with arch suffix are not used in charts or operator files. Support for pushing to quay is available if using `docker/build-push-action`

.github/workflows/build.yml
```diff
@@ -44,9 +44,64 @@ jobs:
...
+
+      - name: Build and push the ansible image
+        uses: docker/build-push-action@v2
+        with:
+          file: ./e2e-tests/Dockerfile
+          push: true
+          load: false
+          platforms: linux/amd64
+          tags: |
+            ${{ secrets.IMAGE_ORG}}/lvm-localpv-e2e:ci
```
akhilerm, line 100: This can be made an env. So that we can default to openebs. Ref https://github.com/openebs/node-disk-manager/blob/master/.github/workflows/build.yml#L63

akhilerm (approved):
LGTM

shubham14bajpai (approved):

PR #28, opened 2021-03-03
feat(provisioning): add support for multiple vg to use for provisioning

### LINKED ISSUE #17: support for multiple volume group within a given node
[A description of the current limitation/problem/challenge that you are experiencing.]
Do we have any plans to support multiple volume groups within a storage class? As of now, I see there is `volgroup` parameter is being used by controller that one specifies in the storage class. But I think there is no way to specify multiple volume groups exists in a node to provision the logical volumes. It may be required if one is using multiple disks within a node of same type. Rather than creating single vg group across all disks of same type which introduces single point of failure with degradation of one disk affecting all lvs (lvm stripping), we would like to use multiple vgs.

[A clear and concise description of what you want to happen.]
Instead of understanding a single parameter `volgroup`, we can add an ability to specify a volume group name regex to use for provisioning the logical volume. In the node agent side, we can first list all vgs available on the node matching the regex & capacity requirement and then choose some vg (based on say with some configurable bin packing algo) to provision the lv. One simple bin-packing algo could be to choose vg having smallest size available to fit the required lv size.

**Environment:**
- LVM Driver version

iyashu:
See #17 for more details.

**What this PR does?**:
Since the `volgroup` param now represents the regex, controller pass the same to node plugin by setting up an additional field called `VolGroupRegex` in `lvmvolume` resource. Node plugin controller will then chose a vg matching the provided regex and sets up the `VolGroup` field under `lvmvolume`. If there are multiple vgs matching the regex, node plugin controller will choose the one having minimum free space (bin packing) available to accommodate that volume capacity.

Consider a k8s cluster having 4 nodes with vgs [lvmvg-a, lvmvg-b, lvmvg-a, xyzvg]. Configure the storage class by setting `volgroup` parameter as `lvmvg*` (regex denoting a lvmvg prefix). After creating a stateful set of size 4, we'll see each pvc gets scheduled on first 3 nodes (since xyzvg doesn't matches the volgroup regex).

This pull request is dependent on #21. So, we need to close that first before closing this.

**Checklist:**
- [x] Fixes #17 
- [ ] Commit has unit tests


pkg/driver/controller.go
```diff
@@ -263,12 +262,13 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
...
 	owner := selected[0]
-	klog.Infof("scheduling the volume %s/%s on node %s", params.VolumeGroup, volName, owner)
+	klog.Infof("scheduling the volume %s/%s on node %s",
+		params.VgRegex.String(), volName, owner)
 
 	volObj, err := volbuilder.NewBuilder().
 		WithName(volName).
 		WithCapacity(capacity).
-		WithVolGroup(params.VolumeGroup).
+		WithVolGroupRegex(params.VgRegex.String()).
```
pawanpraka1, line 271: this should not work, can you try the storage class without volgroup? Have you tested it with the volgroupregx only in the storageclass?

pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
```diff
@@ -62,9 +62,14 @@ type VolumeInfo struct {
 
 	// VolGroup specifies the name of the volume group where the volume has been created.
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:MinLength=1
```
pawanpraka1, line 65: remove the required option and add omitempty, `json:"volGroup,omitempty"`.

pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
```diff
@@ -62,9 +62,14 @@ type VolumeInfo struct {
...
 	VolGroup string `json:"volGroup"`
 
+	// VolGroupRegex specifies the regex to choose volume groups where volume
+	// needs to be created.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	VolGroupRegex string `json:"volGroupRegex"`
+
 	// Capacity of the volume
 	// +kubebuilder:validation:Required
```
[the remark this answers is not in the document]
pawanpraka1: never mind, we need to store which volgroup we used to create the volume, having one extra field is fine.

pkg/driver/controller.go
```diff
@@ -250,7 +249,7 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
 		}
 	}
 
-	nmap, err := getNodeMap(params.Scheduler, params.VolumeGroup)
+	nmap, err := getNodeMap(params.Scheduler, params.VgRegex)
```
pawanpraka1, line 252: We should using VolumeGroup is VolRegex is empty here. what will happen if someone upgraded and having storageclass with volgroup only.

pkg/driver/params.go
```diff
@@ -17,15 +17,18 @@
...
+	"regexp"
+
 	"github.com/openebs/lib-csi/pkg/common/helpers"
 )
 
 // VolumeParams holds collection of supported settings that can
 // be configured in storage class.
 type VolumeParams struct {
-	// VolumeGroup specifies vg name to use for
+	// VolumeGroup specifies vg regex to use for
```
pawanpraka1, line 29: nit: VolRegex

pkg/lvm/lvm_util.go
```diff
@@ -130,6 +130,11 @@ func CreateVolume(vol *apis.LVMVolume) error {
 
 // DestroyVolume deletes the lvm volume
 func DestroyVolume(vol *apis.LVMVolume) error {
+	if vol.Spec.VolGroup == "" {
```
pawanpraka1, line 133: can it happen?

pkg/driver/params.go
```diff
@@ -49,7 +52,11 @@ func NewVolumeParams(m map[string]string) (*VolumeParams, error) {
 	// parameter keys (not values!) are all lowercase, keys may safely be forced
 	// to the lower case.
 	m = helpers.GetCaseInsensitiveMap(&m)
-	params.VolumeGroup = m["volgroup"]
+
+	var err error
+	if params.VgRegex, err = regexp.Compile(m["volgroup"]); err != nil {
```
pawanpraka1, line 57: shouldn't we use volgroupregex here? why we are using volgroup key here.

iyashu: With multi-pool support, I think it's not required to introduce another storage class parameter. Here we are using the same existing `volgroup` parameter, but with slight different semantics i.e now the value will be treated as regex. If the values doesn't contains any special regex characters, then existing semantics remain same (i.e exact match).

pkg/driver/controller.go
iyashu: Here we not introducing any other storage class param, but rather using the same volgroup param. See the other related comment below.

pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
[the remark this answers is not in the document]
iyashu: Right. Ideally, we should have moved it under the status field but it'll result in larger refactoring exercise.

pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
iyashu: I see `VolumeInfo` struct is also being used in LVMSnapShot crd. I was not sure if removing the required validation has any effect on it. Can you help me to understand what required validation does for string type? Does api-server check for string being non-empty in case of required? I'm not able to find any official documentation about the same.

pkg/lvm/lvm_util.go
iyashu: May be in one corner scenario where deletion of lvmvol happens just after creation but before provisioning. e.g our creation workflow gets failed due to some reason & gets retried after some time but deletion event gets triggered in between.

pawanpraka1:
@iyashu let me know what do you think about this approach :

1. have a same paramater `volgroup` in the storageclass which can be used for both regx and for providing volgroup name as well
2. now same parameter will be used in the LVMVolume CRD (VolGroup) to pass to the node daemonset
3. node daemonset will find the volume group on the node based on whether it is regx or full name
4. once it finds the best volume group for provisioning the volume, it will update LVMVolume CRD with that volume group name, also we can update the regx in the annotation on this LVMVolume CRD for debug/other purpose. 
5. now we don't need to have one extra feild and same volGroup can be used to provide the regx also.

iyashu:
> 4. once it finds the best volume group for provisioning the volume, it will update LVMVolume CRD with that volume group name(update the VolGroup paramenter), also we can update the regx in the annotation on this LVMVolume CRD for debug/other purpose.

@pawanpraka1 there are some failure recovery scenarios missing in above approach. Say, we provisioned the volume as per the VolGroup regex attribute, but failed to update the same (say kube-api-server failing intermittently) in api-server. In that case, controller retry may end up creating new lv on some other vg. Although, we can check if that lv is being created already on any vg or not by parsing lvs output, but it'll unnecessary add complexity to maintain idempotency in creation as well as deletion workflow of control loop. Also, putting regex in the annotation may not be a right k8s crd development approach since annotations are not designed for storing the spec, but just for metadata. Here volRegex is the spec (intend) from controller to the node agent to provision the volume rather some additional metadata.

It may not be good idea to represent different semantics with same attribute in the crd. First we are considering volGroup as intend (regex) for provisioning the volume and after provisioned we are considering it as output. Since this attribute is used across the project (including the snapshotting LVMSnapshot crd), It'll just confuse us wherever we are going to reference like whether it should be treated as intend or output & one needs to check the status field every time to assert it.

With current implementation, ideally i see we can move volGroup field from spec to status in future once we do some refactoring specially de-coupling snapshot crd with lvm volume spec. But till that we can just mentally think of it as part of status.

pawanpraka1:
>  In that case, controller retry may end up creating new lv on some other vg.

We should handle this case as this can happen with the current approach also when it failed to update the status to Ready.

We need to keep the storageclass parameters in sync with the lvm volume CRDs. I agree we should not have  different semantics for the same argument and we can solve that by passing regx in the annotaton as this is a meta info for lvm volume and keep the volgroup as empty until node agent fills that info.

iyashu:
@pawanpraka1 updated the pull request as per our discussion today, ptal.

pawanpraka1:
@iyashu can you update the storageclass parameter name to vgpattern?

iyashu:
> @iyashu can you update the storageclass parameter name to vgpattern?

done.

pkg/mgmt/volume/volume.go
```diff
@@ -96,14 +99,80 @@ func (c *VolController) syncVol(vol *apis.LVMVolume) error {
...
+	} else {
+		for _, vg := range vgs {
+			// first update volGroup field in lvm volume resource for ensuring
+			// idempotency and avoiding volume leaks during crash.
+			if vol, err = lvm.UpdateVolGroup(vol, vg.Name); err != nil {
+				klog.Errorf("failed to update volGroup to %v: %v", vg.Name, err)
+				return err
+			}
+			if err = lvm.CreateVolume(vol); err == nil {
+				return lvm.UpdateVolInfo(vol, lvm.LVMStatusReady)
```
pawanpraka1, line 129: there could be a leak here if UpdateVolInfo fails, and it pick a new vg for creating the volume. We need to fix it @iyashu

iyashu: I think there will not be any leak here. See we are setting up the vg name first in the spec using `lvm.UpdateVolGroup(vol, vg.Name)` at line 124 above before creating the lv. Lets say `UpdateVolInfo` fails here, in the next sync iteration, you can see we first check if the vg is set in the spec and accordingly call lvm.CreateVolume with same vg and UpdateVolInfo (see the line 104-109). As a result, we'll keep on retrying the `UpdateVolInfo` if it keeps on failing with backoff set by workqueue.

pawanpraka1: sounds good. Thanks @iyashu

pawanpraka1 (approved):
looks good.

PR #44, opened 2021-04-11
fix(path): fixed the path for node-power-failure test case directory

w3aman:

- Fixed the directory path to node-power-failure e2e-test case.
it should be `node_power_failure` but it was `node_failure`


nsathyaseelan (approved):
lgtm

nsathyaseelan (approved):
lgtm

PR #46, opened 2021-04-15
chore(readme): Update readme with minimum required k8s version

w3aman:

- Update readme with minimum required k8s version
- Update storage class parameters


pawanpraka1 (approved):
looks good.

PR #55, opened 2021-04-23
fix(provision): prevent volume deletion if snapshot exists

akhilerm:
When an LVM volume  with active snapshot is deleted, both volume and snapshot is removed from the machine, but the resources corresponding to the snapshot still exists in the cluster. This stale snapshot resources will cause the controller to error out when we try to delete these resources as the actual snapshot does not exist.

**What this PR does?**:
- error out the DeleteVolume() controller gRPC call if the volume has active snapshots

No

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Has the change log section been updated?
- [ ] Commit has unit tests
- [x] Commit has integration tests

### DIFF
--- a/tests/pv/build.go
+++ b/tests/pv/build.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The OpenEBS Authors
+Copyright 2021 The OpenEBS Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/tests/pv/build.go
+++ b/tests/pv/build.go
@@ -1,0 +1,175 @@
+/*
+Copyright 2019 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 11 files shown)


akhilerm:
@pawanpraka1  I have added the check for snapshot in the controller DeleteVolume(), but the PVC is getting deleted even though DeleteVolune() errors out. How can we prevent the PVC deletion?

pawanpraka1:
@akhilerm, yes, csi-provisioner does not honour the error returned from delete volume call. Let us put this question in the CSI channel. Here what we can do is, we can error out the delete request at the agent side.

akhilerm:
>  what we can do is, we can error out the delete request at the agent side.

@pawanpraka1 If we error out at agent side, will it prevent the PVC deletion?

pawanpraka1:
> If we error out at agent side, will it prevent the PVC deletion?

no, PVC will be deleted any way. Here we can only control whether the actual volume will be deleted or not at the agent side.

tests/pv/build.go
```diff
@@ -0,0 +1,175 @@
+/*
+Copyright 2019 The OpenEBS Authors
```
pawanpraka1, line 2: 2021

tests/pv/buildlist.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2019 The OpenEBS Authors
```
pawanpraka1, line 2: 2021?

tests/utils.go
```diff
@@ -457,3 +455,52 @@ func deletePVC(pvcname string) {
...
+
+	shouldPVExist := gomega.BeFalse()
+	if shouldExist {
+		shouldPVExist = gomega.BeTrue()
+	}
+
+	ginkgo.By("verifying PV for PVC exists or not")
+	var matchingPVName string
+	matchingPVExists := false
+	for _, pv := range pvList.Items {
```
pawanpraka1, line 475: this looks ok. Can't we query the pv before deleting the pvc and then delete the pvc and verify the pv?

akhilerm: yes, we can . It will just verify that the PV is present for the PVC.  We are already checking if the PVC is in Bound state, so it will be just an additional check to the same.

pawanpraka1 (approved):
looks good.

PR #76, opened 2021-05-21
fix(prov): add controller to avoid potential volume leaks

iyashu:
Here is the sequence of events that can produce a volume leak:

1. create a new pvc with late binding.
2. provision the volume in your csi driver `CreateVolume` as usual, but return artificial unsuccessful response (non-ok grpc code) to the csi external provisioner even though you are able to successfully provisioned the volume.
3. delete the pvc that you created earlier and wait for it till it gets completely removed.
4. terminate the csi controller plugin pod.
5. even after restarting the csi controller plugin pod, you'll see that underlying volume (lvm volume resource in our case) still exists without any pvc or pv associated with it thus leaking the underlying storage infra structure.

It is a known limitation of csi spec itself where CO can't invoke delete volume method without knowing the volume id (from CreateVolume response) as the create volume rpc keeps returning the error response. With restarting external provisioner, it loses its in-memory state of the pending create volume and stops retrying as the pvc resource has been deleted.

**What this PR does?**:
A separate controller watching pvc resources and add `csi-leak-protection` finalizer to delay its deletion until it make sure that the underlying storage volume gets garbage collected or the corresponding pvc gets bound.

No.

Manually failing the CreateVolume rpc calls from external provisioner even after provisioning the volume. On deleting the corresponding pvc, we can see that associated lvm volume also gets deleted.

I've tried making the `LeakProtectionController` keeping general csi driver plugins implementations in mind, so that it can be abstracted in separate common package (may be lib-csi in future) and can be used across all the csi driver with minimal changes at driver side.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -25,6 +25,8 @@ import (
 	k8sapi "github.com/openebs/lib-csi/pkg/client/k8s"
 	clientset "github.com/openebs/lvm-localpv/pkg/generated/clientset/internalclientset"
 	informers "github.com/openebs/lvm-localpv/pkg/generated/informer/externalversions"
+	"github.com/openebs/lvm-localpv/pkg/mgmt/csipv"
+	corev1 "k8s.io/api/core/v1"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
@@ -67,6 +69,8 @@ type controller struct {
 
 	k8sNodeInformer cache.SharedIndexInformer
 	lvmNodeInformer cache.SharedIndexInformer
+
+	leakProtection *csipv.LeakProtectionController
 }
 
 // NewController returns a new instance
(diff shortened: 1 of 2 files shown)


iyashu:
Gentle Reminder! 
@prateekpandey14 @akhilerm @shubham14bajpai

shubham14bajpai (approved):
LGTM

pkg/driver/controller.go
```diff
@@ -339,49 +368,38 @@ func (cs *controller) DeleteVolume(
...
-	if err != nil {
-		return nil, errors.Wrapf(
-			err,
-			"failed to handle delete volume request for {%s}",
-			volumeID,
-		)
+	if vol.GetDeletionTimestamp() == nil {
+		if err = lvm.DeleteVolume(volumeID); err != nil {
+			return errors.Wrapf(err,
+				"failed to handle delete volume request for {%s}", vol.GetName())
```
prateekpandey14, line 395: any sepcific reason to use `vol.GetName()` here instead of  just `volumeID` here in error,similar to above error ?

pkg/driver/controller.go
```diff
@@ -339,49 +368,38 @@ func (cs *controller) DeleteVolume(
...
-
-	// Delete the corresponding ZV CR
-	err = lvm.DeleteVolume(volumeID)
-	if err != nil {
-		return nil, errors.Wrapf(
-			err,
-			"failed to handle delete volume request for {%s}",
-			volumeID,
-		)
+	if vol.GetDeletionTimestamp() == nil {
```
prateekpandey14, line 392: Is it to avoid the ongoing delete volume call ? if yes can we add the code comment here.

pkg/mgmt/csipv/leak_protection.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	alreadyExists := c.claimsInProgress.Add(key)
+	if alreadyExists {
+		return nil, status.Errorf(codes.Aborted,
+			"csi driver already has volume creation in progress")
+	}
+
+	if err = c.addFinalizer(pvc, volumeName); err != nil {
+		finishCreateVolume() // make sure we clean up on error.
+		return nil, err
```
prateekpandey14, line 322: can we return the func here also instead of calling it here explicitly and then return nil  ?

prateekpandey14 (commented):
looks good, some minor comments though

pkg/driver/controller.go
iyashu: yes! Done.

pkg/driver/controller.go
iyashu: No specific reason. Probably got carried away with refactoring. I'll update it to keep it consistent.

pkg/mgmt/csipv/leak_protection.go
iyashu: It is cause of non-nil error. Essentially, the caller should only consider return value as valid if the error nil. Otherwise, the caller have to check if return value is valid every time regardless of error is nil or not.

akhilerm (approved):
LGTM

prateekpandey14 (approved):
lgtm

pawanpraka1 (approved):
looks good.

pawanpraka1:
Thanks for this PR @iyashu.

PR #86, opened 2021-06-03
feat(charts): set default fstype to ext4 in csi-provisioner

shovanmaity:
Set default fstype to ext4 in csi-provisioner. This will be helpful when fsType is not mention in storageclass.

### DIFF
--- a/deploy/lvm-operator.yaml
+++ b/deploy/lvm-operator.yaml
@@ -1182,6 +1182,7 @@ spec:
             - "--strict-topology"
             - "--leader-election"
             - "--extra-create-metadata=true"
+            - "--default-fstype=ext4"
           env:
             - name: ADDRESS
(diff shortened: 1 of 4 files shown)


deploy/lvm-operator.yaml
```diff
@@ -1182,6 +1182,7 @@ spec:
             - "--strict-topology"
             - "--leader-election"
             - "--extra-create-metadata=true"
+            - "--default-fstype=ext4"
```
pawanpraka1, line 1185: This file is autogenerated, you need to make the same change in deploy/yamls/zfs-driver.yaml

pawanpraka1 (approved):
looks good.

deploy/lvm-operator.yaml
shovanmaity: done

shubham14bajpai (approved):
LGTM

PR #96, opened 2021-06-14
chore(readme): Update readme with lvm driver installation from charts repo

w3aman:

- Update readme with lvm driver installation step from charts repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,10 @@ sudo vgcreate lvmvg /dev/loop0       ## here lvmvg is the volume group name to b
 
 ### Installation
 
-We can install the LVM-LocalPV driver by running the following command.
+We can install the latest release of OpenEBS LVM driver by running the following command.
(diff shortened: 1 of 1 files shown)


pawanpraka1 (approved):
looks good. Can we also add this from ZFS repo 
```
NOTE: For some Kubernetes distributions, the kubelet directory must be changed at all relevant places in the YAML powering the operator (both the openebs-zfs-controller and openebs-zfs-node).

For microk8s, we need to change the kubelet directory to /var/snap/microk8s/common/var/lib/kubelet/, we need to replace /var/lib/kubelet/ with /var/snap/microk8s/common/var/lib/kubelet/ at all the places in the operator yaml and then we can apply it on microk8s.

For k0s, the default directory (/var/lib/kubelet) should be changed to /var/lib/k0s/kubelet.

```

PR #142, opened 2021-08-30
feat(e2e): add e2e-test for raw block volume for lvm-localpv

w3aman:

- This PR adds e2e-test for raw block volume for lvm-localpv


nsathyaseelan (approved):
lgtm

PR #146, opened 2021-09-07
feat(helm): Disable leader-elections if replicas=1, make logLevel configurable

Quarky9:
- allows to optionally reduce log noise
- reduces load on controllers / etcd for single replica deployments

**What this PR does?**:
Disable leader-election for the controllers if replicas = 1

No.


pawanpraka1:
@Quarky9 you need to update the chart version also here https://github.com/openebs/lvm-localpv/blob/master/deploy/helm/charts/Chart.yaml#L4.

kmova (approved):

PR #148, opened 2021-09-13
refact(e2e): Cleanup csi-driver CR before applying updated operator yaml

w3aman:

- Due to Change in CSI_driver spec, we need to delete csi-driver CR before apply the updated operator yaml, otherwise csi-driver will no get updated as of immutable spec fields. 

```
The CSIDriver "local.csi.openebs.io" is invalid: spec.storageCapacity: Invalid value: true: field is immutable

```


nsathyaseelan (approved):
lgtm

PR #178, opened 2022-02-25
feat(scheduler): adding SpaceWeighted scheduler

### LINKED ISSUE #104: LVM-LocalPV should pick the node where enough space is available for immediate binding
Currently for the immediate binding case we have two scheduling logic 
1. Volumeweighted, here we pick the node where less number of volumes are provisioned
2. CapacityWeighted here we pick the node where size of all the provisioned volumes are less

These 2 are not sufficient as there is a possibility of getting a node where enough space is not availabe as none of above scheduling logic look for the free space available in the volumegroup.

Now that since we have lvmnode CR which has all the information about the volume group. We can use that to find the best node. We need to create a node map (https://github.com/openebs/lvm-localpv/blob/d4c4bebc81c9f6f955c0d03f0cc2fb991c6cf374/pkg/driver/schd_helper.go#L96) of free space in the volume group. We should make this as default option for scheduling.

The schedule currently picks the node which one less weighted(https://github.com/openebs/lib-csi/blob/develop/pkg/scheduler/scheduler.go#L81) in the map, we need to modify it to pick the highest weighted node as here we need to pick the node where free space is more.

pawanpraka1:
fixes: https://github.com/openebs/lvm-localpv/issues/104

adding a new scheduler logic which selects the node
which has max free space available. This is better scheduling
algorithm than the volumeweighted and capacity weighted where we
don't check for the free space available on the node.

**Release note**
Making it a default scheduler, so, now onwards, for immediate binding case, the controller will pick the node which has largest free space available for provisioning the volumes.

### DIFF
--- a/pkg/driver/schd_helper.go
+++ b/pkg/driver/schd_helper.go
@@ -34,10 +34,10 @@ const (
 	VolumeWeighted = "VolumeWeighted"
 
 	// pick the node where total provisioned volumes have occupied less capacity from the given volume group
-	// this will be the default scheduler when none provided
 	CapacityWeighted = "CapacityWeighted"
 
 	// pick the node which is less loaded space wise
+	// this will be the default scheduler when none provided
 	SpaceWeightedMap = "SpaceWeighted"
 )
 
@@ -123,6 +123,7 @@ func getSpaceWeightedMap(re *regexp.Regexp) (map[string]int64, error) {
 		}
 		if maxFree > 0 {
 			// converting to SpaceWeighted by subtracting it with MaxInt64
+			// as the node which has max free space available is less loaded.
 			nmap[node.Name] = math.MaxInt64 - maxFree
 		}
 	}
--- a/pkg/driver/schd_helper.go
+++ b/pkg/driver/schd_helper.go
@@ -17,11 +17,13 @@ limitations under the License.
 package driver
 
 import (
+	"math"
 	"regexp"
 	"strconv"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/openebs/lvm-localpv/pkg/builder/nodebuilder"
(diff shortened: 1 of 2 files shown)


pkg/driver/schd_helper.go
```diff
@@ -34,6 +36,9 @@ const (
 	// pick the node where total provisioned volumes have occupied less capacity from the given volume group
 	// this will be the default scheduler when none provided
```
akhilerm: remove this comment and add it for spaceweighted as it is the default

pkg/driver/schd_helper.go
```diff
@@ -92,14 +97,49 @@ func getCapacityWeightedMap(re *regexp.Regexp) (map[string]int64, error) {
...
+			if re.MatchString(vg.Name) {
+				freeCapacity := vg.Free.Value()
+				if maxFree < freeCapacity {
+					maxFree = freeCapacity
+				}
+			}
+		}
+		if maxFree > 0 {
+			// converting to SpaceWeighted by subtracting it with MaxInt64
+			nmap[node.Name] = math.MaxInt64 - maxFree
```
akhilerm, line 127: Should this be node.Name or OwnerNodeID, as we use node in capacity weighted and volume weighted.

akhilerm: It will be better to add the comment on why we are subtracting and using. It takes sometime to understand the logic from the code directly.

akhilerm (commented):
@pawanpraka1 Have a few nitpick comments

pkg/driver/schd_helper.go
pawanpraka1: it should be node.Name. capacity weighted and volume weighted are volume bases algorithm. added comment.

akhilerm (approved):
LGTM.

pawanpraka1:
@w3aman please verify this feature.

PR #183, opened 2022-04-06
chore(snapshot): adding the design doc for mountable snapshots

wowditi:
## Pull Request template

**What this PR does?**: It adds a design for a new feature

**Does this PR require any upgrade changes?**: No

In the drawbacks section I posed the question of whether we should maybe use annotation to indicate that we are mounting a snapshot instead of creating a new PV as a clone of the snapshot. I'd like a second opinion on whether this is a good idea or what a good alternative could be (for example, we could create a new api-resource although I don't think that would be such a good idea).
Also a PVC requires the Storage to be set, however, it is a little bit useless in this case especially when mounting snapshots. Any suggestions on handling that better would be appreciated.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/design/lvm/mountable_snapshot.md
+++ b/design/lvm/mountable_snapshot.md
@@ -188,4 +188,4 @@ All testcases mentioned in [Test Plan](#test-plan) section need to be automated
 ## Drawbacks
 
 - As far as I understand it, normally when creating a PersistentVolumeClaim with a snapshot as datasource it would clone the Snapshot into an actual volume, so we'd be diverging from this behaviour.
-  - We could use an annotation to specify the behaviour that we are implementing here so that the clone behaviour can be added at a later time?
+  - We could use an annotation (e.g. `local.csi.openebs.io/volume-type: mounted-snapshot`) applied to the PVC to specify the behaviour that we are implementing here. In the `CreateVolume` function of the [controller.go](../../pkg/driver/controller.go) we could then check whether this annotation is set and only then mount the snapshot. When the annotation is not set or set to `local.csi.openebs.io/volume-type: clone` the code would create a clone instead (which for now would `return nil, status.Error(codes.Unimplemented, "")`) until someone implements the logic to clone a snapshot. 
--- a/design/lvm/mountable_snapshot.md
+++ b/design/lvm/mountable_snapshot.md
@@ -162,6 +162,7 @@ The changes below are the primary modules that need to be modified, however, the
 
 - In the [controller.go](../../pkg/driver/controller.go) the code path in the `CreateVolume` function for the `controller` type that occurs when contentSource.GetSnapshot() is not `nil` needs to be implemented. When this path is triggered it should return the correct `volName`, `topology` and `cntx`.  
 - In the [agent.go](../../pkg/driver/agent.go) the `NodePublishVolume` function for the `node` type needs to be changed such that it checks whether the `volName` is a snapshot and if so mount the snapshot to the specified location.
+  - Finding the snapshot that is to be used can be accomplished by taking and VolumeId and removing the `snapshot-` prefix.
   - This also needs to change the write access to `rw` by using the lvchange command when the PersistentVolumeClaim specified the AccessMode as ReadWrite. Note that we do not need to change it back to read only, since we can limit the permissions of future mounts to read only by using the `MountOptions`. 
     - Alternatively we could make the Snapshots writeable by default
 - In the [agent.go](../../pkg/driver/agent.go) the `NodeUnpublishVolume` function for the `node` type needs to be changed such that it unmounts the snapshot.
--- a/design/lvm/mountable_snapshot.md
+++ b/design/lvm/mountable_snapshot.md
@@ -1,0 +1,191 @@
+---
+title: Snapshot Mount Support for LVM-LocalPV
+authors:
+  - "@wowditi"
+owners:
+  - "@kmova"
+creation-date: 2022-03-31
+last-updated: 2022-03-31
+status: Request for Comment
+---
+
+# Snapshot Mount Support for LVM-LocalPV
+
+## Table of Contents
+
+* [Table of Contents](#table-of-contents)
+* [Summary](#summary)
+* [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+* [Implementation Details](#implementation-details)
+* [Test Plan](#test-plan)
+* [Graduation Criteria](#graduation-criteria)
+* [Drawbacks](#drawbacks)
+
+## Summary
+
+LVM Snapshots are space efficient and quick point in time copies of lvm volumes. It consumes the space only when changes are made to the source logical volume. For testing purposes it can be beneficial to start a new application, for example a database, using such a snapshot as the backing storage volume. This allows for creating an exact replica of the application at the moment the snapshot was taken (assuming the state is entirely depedent on the filesystem), which can used to debug issues that occured in production or test how a new version of the application (or an external application) would affect the data.
+
+## Motivation
+
+### Goals
+
+- user should be able to mount snapshots
+- user should be able to mount thick and thin snapshots
+- user should be able to mount snapshots as read only and copy on write 
+
+### Non-Goals
+
+- Creating clones from Snapshots
+- restore of a snapshot
+
+
+## Proposal
+
+To mount a k8s lvmpv snapshot we need to create a persistent volume claim that references the snapshot as datasource:
+
+```
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: csi-lvmpvc-snap
+spec:
+  storageClassName: openebs-lvmpv
+  dataSource:
+    name: lvmpv-snap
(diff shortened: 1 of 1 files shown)


design/lvm/mountable_snapshot.md
```diff
@@ -0,0 +1,191 @@
...
+- Verify that creating a PersistentVolumeClaim for a non existing VolumeSnapshot fails
+
+## Graduation Criteria
+
+All testcases mentioned in [Test Plan](#test-plan) section need to be automated
+
+## Drawbacks
+
+- As far as I understand it, normally when creating a PersistentVolumeClaim with a snapshot as datasource it would clone the Snapshot into an actual volume, so we'd be diverging from this behaviour.
+  - We could use an annotation to specify the behaviour that we are implementing here so that the clone behaviour can be added at a later time?
```
pawanpraka1: can you mention the annotation and also how we are going to use that?

design/lvm/mountable_snapshot.md
```diff
@@ -0,0 +1,191 @@
...
+The changes below are the primary modules that need to be modified, however, there are some additional utility modules that will need to be modified to support these changes. Most likely the following files will need some additions to support snapshots: [kubernetes.go](../../pkg/builder/volbuilder/kubernetes.go) (in order to implement a `Kubeclient.GetSnapshot` function), [lvm_util.go](../../pkg/lvm/lvm_util.go) (in order to create a function that changes the snapshot write access) and [mount.go](../../pkg/lvm/mount.go) (in order to support mounting/unmounting of snapshots).
+
+- In the [controller.go](../../pkg/driver/controller.go) the code path in the `CreateVolume` function for the `controller` type that occurs when contentSource.GetSnapshot() is not `nil` needs to be implemented. When this path is triggered it should return the correct `volName`, `topology` and `cntx`.  
+- In the [agent.go](../../pkg/driver/agent.go) the `NodePublishVolume` function for the `node` type needs to be changed such that it checks whether the `volName` is a snapshot and if so mount the snapshot to the specified location.
```
pawanpraka1, line 164: can you describe how we are going to find which snapshot we have to mount, how do we find that?

Ab-hishek, line 164: Why do we need to know whether the `volName` is a snapshot? We are creating a volume out of the snapshot and then simply we want to mount that newly created volume. Am I interpreting it right or am I missing something?

design/lvm/mountable_snapshot.md
```diff
@@ -0,0 +1,191 @@
...
+    storage: 10Gi
+  phase: Bound
+```
+
+## Implementation Details
+
+In order to implement this only existing code needs to be changed, there is no need for an entirely new control flow.
+The changes below are the primary modules that need to be modified, however, there are some additional utility modules that will need to be modified to support these changes. Most likely the following files will need some additions to support snapshots: [kubernetes.go](../../pkg/builder/volbuilder/kubernetes.go) (in order to implement a `Kubeclient.GetSnapshot` function), [lvm_util.go](../../pkg/lvm/lvm_util.go) (in order to create a function that changes the snapshot write access) and [mount.go](../../pkg/lvm/mount.go) (in order to support mounting/unmounting of snapshots).
+
+- In the [controller.go](../../pkg/driver/controller.go) the code path in the `CreateVolume` function for the `controller` type that occurs when contentSource.GetSnapshot() is not `nil` needs to be implemented. When this path is triggered it should return the correct `volName`, `topology` and `cntx`.
```
Ab-hishek, line 163: Do we have anything specific in mind on how to implement this?

design/lvm/mountable_snapshot.md
```diff
@@ -0,0 +1,191 @@
...
+
+To mount a k8s lvmpv snapshot we need to create a persistent volume claim that references the snapshot as datasource:
+
+```
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: csi-lvmpvc-snap
+spec:
+  storageClassName: openebs-lvmpv
```
Ab-hishek, line 55: Is storage class required to be of `local.csi.openebs.io` provisioner type? Can we have different storage class consuming the lvm snapshots?

design/lvm/mountable_snapshot.md
wowditi: I've added some additional details to the document. Basically we annotate the PVC to indicate whether it needs to clone the snapshot or mount it.

design/lvm/mountable_snapshot.md
wowditi: I've added some additional details in the doc as to how this can be done. But it's rather straightforward, the VolumeId (which is should be provided to this function in the request) should be of the format `snapshot-<uuid>` (or possibly `pvc-<original-uuid>@snapshot-<snapshot-uuid>`) where the uuid is also the name of the snapshot on the host.

wowditi: I figured we'd need to call different functions so that we end up with an `*apis.LVMSnapshot` object instead of a `*apis.LVMVolume` object, but that might not even be needed. I suppose that's mainly an implementation detail and this should simply say that it mount the snapshot.

design/lvm/mountable_snapshot.md
wowditi: As a matter of fact yes, I created a quick PoC a while back to see if this would even be feasible. 
The first step is determining the volume name, which can be parsed from the SnapshotId (`contentSource.GetSnapshot().GetSnapshotId()`). I'd suggest passing `snapshot-<uuid>` as the volume name, but the entire snapshot_id would also work. 
We'd also need to create a function that retrieves the `*lvmapi.LVMSnapshot` object based on the request (instead of the `*lvmapi.LVMVolume` object we get after creating a new LVM Volume. The process is quite similar to that of building a regular volume (using the `volbuilder.NewBuilder()` function). The only information that is not already present is that of the owner, but we can retrieve this by using the existing `lvm.GetLVMSnapshot` function.
With this information creating the `topology` and `cntx` is exactly the same as before only using the `*lvmapi.LVMSnapshot` instead of the `*lvmapi.LVMVolume`.

This feels too detailed to be in a design document and more an implementation specific part, but if you disagree I can add it to the design doc.

design/lvm/mountable_snapshot.md
wowditi: Since the `dataSource` is provided separately I don't think it's required. I'm not entirely sure what's involved with creating a new provisioner type, but it would be a good solution to the downside that I mentioned in the design.

Ab-hishek (changes requested):
Hi @wowditi. This document looks good. Just a minor comment. Since, this is a design document we really don't need to mention the exact files or function names where the changes need to be done. We can simply mention the goals, non-goals, design change/enhancement, any diagrams if needed. Therefore can you make the necessary changes for it.

Abhinandan-Purkait:
@wowditi Can you please provide an update on the plans for taking this forward? Are you willing to work on this?

orville-wright:
Hi @wowditi I run Product Mgmt for the OpenEBS project.
You've put a lot of amazingly impressive & detailed work into your PR and it received a bunch of good reviews from our ENG team. Overall very positive. 
LVM-localPV is now getting a major focus in the OpenEBS product as it's being unified into the core platform as a central component, rather than being managed as a separate external Data-Engine. 
Additionally, LVM-LocalPV now has ~50,000 Daily Active Users and +150,000 product installations. So it's now a very well validated part of the platform. Hence we're are unifying it into the main product. 

We'd really like to work on developing your PR and integrate the design into the new unified product. 

Are you still interested and willing to support your PR?

wowditi:
Hi, 
Sorry I forgot about this since my company started using the ZFS variant since it already had the features we need. 
I do not have a lot of spare time currently, but I could probably help since it wasn't that much work from what I remember. 

About the design document, the implementation details are indeed very technical currently and I can make those a bit more high level.
After using the ZFS localpv I also noticed that this is basically the same behaviour as what they refer to as cloning, should I rewrite the document to call it cloning and remove the drawback?

PS: I should probably mention that I mainly code in python and not Golang, so it might not be the most beautiful code.

dsharma-dc:
I see at least a few things that aren't talked about in this document:

1. There is an idea in the document that if the PVC for snapshot is created with a capacity(resource)  - `we could extend the lvm if the defined storage size is larger then that of the snapshot (with a maximum of the size of the original volume)`. If the snapshot is thick and is extended to a cap of original LV size, it'll become `invalid` in view of LVM. How do we deal with that **during** PVC provisioning.
2. If say there is a volume `V1` of size `S`, a snapshot of `V1` is created of size 50% i.e. `S/2`. Now a PVC `P` is provisioned with snapshot. If the snapshot is configured to auto-extend, the snapshot size at some point becomes larger than size of `P`. How does that affect the pvc `P` ?
3. If a mounted snapshot becomes invalid (100% of original LV size), how does that impact the pvc which is bound to it?

dsharma-dc:
This hasn't yet progressed based on last set of clarifying questions. Keeping a track of it.

rohan2794:
OpenEBS LVM supports snapshot cloning, so mountable snapshots are not required. Closing this PR.

Issue #163, opened 2021-10-17
openebs-lvm-plugin is busylooping on systems without dm-snapshot kernel module loaded


flokli:
**What steps did you take and what happened:**

I tried to snapshot a volume created by lvm-localpv.

I created a [Default SnapshotClass Without SnapSize Parameter](https://github.com/openebs/lvm-localpv/blob/51707f7f3fa4ab643feabde7a459743bfc70b33e/docs/snapshot.md#default-snapshotclass-without-snapsize-parameter), and then a `VolumeSnapshot` resource:

```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: lvm-localpv-snap
spec:
  volumeSnapshotClassName: lvmpv-snapclass
  source:
    persistentVolumeClaimName: datadir-redpanda-0
```

**What did you expect to happen:**
Snapshot to be created.

* `kubectl logs -f openebs-lvm-controller-0 -n kube-system -c openebs-lvm-plugin`

```
I1017 09:26:48.057768       1 grpc.go:72] GRPC call: /csi.v1.Controller/CreateSnapshot requests {"name":"snapshot-ccd9b76d-b16f-4a17-9809-0f769c0bc2f8","source_volume_id":"pvc-d293d56b-45da-412d-ab62-fec20652e71b"}
I1017 09:26:48.057937       1 controller.go:572] CreateSnapshot volume snapshot-ccd9b76d-b16f-4a17-9809-0f769c0bc2f8 for pvc-d293d56b-45da-412d-ab62-fec20652e71b
I1017 09:26:48.068293       1 grpc.go:81] GRPC response: {"snapshot":{"creation_time":{"seconds":1634462808},"snapshot_id":"pvc-d293d56b-45da-412d-ab62-fec20652e71b@snapshot-ccd9b76d-b16f-4a17-9809-0f769c0bc2f8","source_volume_id":"pvc-d293d56b-45da-412d-ab62-fec20652e71b"}}
I1017 09:26:48.650574       1 grpc.go:72] GRPC call: /csi.v1.Identity/GetPluginInfo requests {}
I1017 09:26:48.650691       1 grpc.go:81] GRPC response: {"name":"local.csi.openebs.io","vendor_version":"0.8.2"}
[...]

PR #186, opened 2022-04-22
refactor: refactor controller to use dynamic client and use better ratelimiter for controlled exponential backoff

abhilashshetty04:
Signed-off-by: Abhilash Shetty <abhilashshetty@abhilashshetty-mbp1.local>

**What this PR does?**:
1. Replaced codes which implemented code generation and used dynamic client to watch lvmvolume, lvmnode and lvmsnapshot cr
2. Replaced builder pattern by one func implementation which would create controller struct and return.
3. Added additional loggers to help debug event triggers.

Snapshot busylooping Fix:
We have changed the ratelimiter which was used in snapshot controller. This ratelimiter requeues failed items after 5 secs for first 12 attempts. Then objects are requeued after 30 secs. I had unloaded dm-snapshot kernel module to reproduce this. Attached logs from the relevant node plugin below.

[retrylogs.txt](https://github.com/openebs/lvm-localpv/files/8899854/retrylogs.txt)

**Issues:**
- [X] Fixes #<issue number> https://github.com/openebs/lvm-localpv/issues/189
- [X] Fixes #<issue number> https://github.com/openebs/lvm-localpv/issues/163


pkg/mgmt/lvmnode/builder.go
```diff
@@ -17,12 +17,12 @@
 package lvmnode
 
 import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
niladrih: Could you use the import order convention:
standard library packages
external packages
local packages

pkg/mgmt/lvmnode/builder.go
```diff
@@ -67,91 +75,29 @@ type NodeController struct {
 	ownerRef metav1.OwnerReference
 }
 
-// NodeControllerBuilder is the builder object for controller.
```
niladrih, line 70: Why are we removing the builder code?
The builder pattern may be more useful here as the Controller may be extended to different configurations easily, without breaking existing code (i.e. by altering function signature).

pkg/mgmt/lvmnode/lvmnode.go
```diff
@@ -18,6 +18,8 @@ package lvmnode
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
niladrih: Import order should follow convention

pkg/mgmt/lvmnode/lvmnode.go
```diff
@@ -116,9 +123,23 @@ func (c *NodeController) syncNode(namespace string, name string) error {
 	return nil
 }
 
+func (c *NodeController) getStructuredObject(obj interface{}) (*apis.LVMNode, bool) {
+	unstructuredInterface, ok := obj.(*unstructured.Unstructured)
+	if ok {
+		node := &apis.LVMNode{}
+		err := runtimenew.DefaultUnstructuredConverter.FromUnstructured(unstructuredInterface.UnstructuredContent(), &node)
+		if err != nil {
+			fmt.Printf("err %s, While converting unstructured obj to typed object\n", err.Error())
```
niladrih: We're using klog for logging. Use klog.Errorf() instead

pkg/mgmt/snapshot/snapshot.go
```diff
@@ -45,14 +47,20 @@ func (c *SnapController) syncHandler(key string) error {
...
 		return nil
 	}
 	if err != nil {
 		return err
 	}
+	snap := apis.LVMSnapshot{}
+	err = runtimenew.DefaultUnstructuredConverter.FromUnstructured(unstructuredSnap.UnstructuredContent(), &snap)
+	//err = runtime.DefaultUnstructuredConverter.FromUnstructured(Vol.UnstructuredContent(), &vol)
+	if err != nil {
+		fmt.Printf("err %s, While converting unstructured obj to typed object\n", err.Error())
```
niladrih: We're using klog. Use klog.Errorf() instead

pkg/mgmt/snapshot/snapshot.go
```diff
@@ -149,10 +157,24 @@ func (c *SnapController) deleteSnap(obj interface{}) {
...
 	c.enqueueSnap(snap)
 }
 
+func (c *SnapController) getStructuredObject(obj interface{}) (*apis.LVMSnapshot, bool) {
+	unstructuredInterface, ok := obj.(*unstructured.Unstructured)
+	if ok {
+		snap := &apis.LVMSnapshot{}
+		err := runtimenew.DefaultUnstructuredConverter.FromUnstructured(unstructuredInterface.UnstructuredContent(), &snap)
+		if err != nil {
+			fmt.Printf("err %s, While converting unstructured obj to typed object\n", err.Error())
```
niladrih: Use klog, instead of fmt

pkg/mgmt/volume/volume.go
```diff
@@ -49,31 +51,113 @@ func (c *VolController) syncHandler(key string) error {
...
 	}
 	if err != nil {
 		return err
 	}
-	VolCopy := Vol.DeepCopy()
+	vol := &apis.LVMVolume{}
+	err = runtimenew.DefaultUnstructuredConverter.FromUnstructured(unstructuredVol.UnstructuredContent(), &vol)
+	//err = runtime.DefaultUnstructuredConverter.FromUnstructured(Vol.UnstructuredContent(), &vol)
+	if err != nil {
+		fmt.Printf("err %s, While converting unstructured obj to typed object\n", err.Error())
```
niladrih: klog, instead of fmt

niladrih (changes requested):
Requested some changes

niladrih:
@abhilashshetty04 -- All commits require a DCO signature. Commit e405d3c24f248bdd0eed5727f4a666c592ce5e55 is not signed. Also, please add a relevant PR description

pkg/mgmt/volume/volume.go
abhilashshetty04: Incorporated this.

pkg/mgmt/snapshot/snapshot.go
abhilashshetty04: Incorporated this.

pkg/mgmt/lvmnode/builder.go
abhilashshetty04: I was able to change for 2 files. I did try while working on it a week back as well. For some files when i sort it according to best practice and save the file, It gets arranged to the order it was before. Will check why this happens

pkg/mgmt/lvmnode/builder.go
abhilashshetty04: I did check the previous implementation . I felt it to be more complex than it needs to be. In previous implementation as well incase we need more functionality we would need to add new functions and use them in builder call. Now with this where one function builds whole controller we might just have to add new args in function definition . 

This is the standard procedure i have seen across many controller implementation. Reduces multiple function calls as well.

niladrih: Long running list of function arguments affect readability.
Builder pattern makes the Controller build parameters more readable. This is a pattern we have followed throughout this repository for better readability, when building instances of types with many parameters.

**E.g.**
If one tries to extend the code to include another controller with different configuration, and a different use-case, this single function won't be usable for that. That person will have to add another function for his use-case.
With the builder-pattern, we can accomodate most use-cases.

We are not optimising for performance, we are not optimising for lines of code. We are optimising for readability.

pkg/mgmt/lvmnode/builder.go
```diff
@@ -67,91 +75,29 @@ type NodeController struct {
...
-		return nil, err
+	klog.Infof("Creating lvm node controller object")
+	nodeContrller := &NodeController{
+		kubeclientset: kubeClient,
+		clientset:     client,
+		NodeLister:    dynamiclister.New(nodeInformer.GetIndexer(), resource),
+		NodeSynced:    nodeInformer.HasSynced,
+		workqueue:     workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Node"),
+		recorder:      recorder,
+		pollInterval:  60 * time.Second,
```
Ab-hishek, line 100: Is this 60 seconds poll interval intentional? Can we not make it configurable?

pkg/mgmt/lvmnode/lvmnode.go
```diff
@@ -54,13 +56,18 @@ func (c *NodeController) syncHandler(key string) error {
...
 		return err
 	}
+
 	var node *apis.LVMNode
 	if cachedNode != nil {
-		node = cachedNode.DeepCopy()
+		nodeStruct, ok := c.getStructuredObject(cachedNode)
+		if !ok {
+			return err
+		}
```
Ab-hishek: Returning error which may not be related to this is not a good idea. We can return our own error stating that the unstructured object is not what it is intended to be. Something on these lines could be helpful rather than using some previous error stored in the err object.

Ab-hishek: Something like this:
```
runtime.HandleError(fmt.Errorf("Couldn't get node object %#v", cachedNode))
return
```

pkg/mgmt/lvmnode/lvmnode.go
```diff
@@ -142,17 +163,21 @@ func (c *NodeController) updateNode(oldObj, newObj interface{}) {
...
-		if !ok {
-			runtime.HandleError(fmt.Errorf("Couldn't get object from tombstone %#v", obj))
-			return
-		}
-		node, ok = tombstone.Obj.(*apis.LVMNode)
-		if !ok {
-			runtime.HandleError(fmt.Errorf("Tombstone contained object that is not a LVMNode %#v", obj))
-			return
+		unstructuredObj, ok := obj.(*unstructured.Unstructured)
+		if ok {
```
Ab-hishek: What the `!ok` case? Shouldn't we return error in case of this?

pkg/mgmt/lvmnode/start.go
```diff
@@ -47,18 +47,22 @@ func Start(controllerMtx *sync.RWMutex, stopCh <-chan struct{}) error {
...
 	if err != nil {
-		return errors.Wrap(err, "error building openebs clientset")
+		return errors.Wrap(err, "error building dynamic client for lvmnode cr")
 	}
 
 	// setup watch only on node we are interested in.
-	nodeInformerFactory := informers.NewSharedInformerFactoryWithOptions(
-		openebsClient, 0, informers.WithNamespace(lvm.LvmNamespace),
-		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+	nodeInformerFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(openebsClientNew, 5*time.Minute,
```
Ab-hishek, line 57: Is 5 minutes a good time interval for updating the informer factory?

pkg/mgmt/lvmnode/start.go
```diff
@@ -47,18 +47,22 @@ func Start(controllerMtx *sync.RWMutex, stopCh <-chan struct{}) error {
...
+		lvm.LvmNamespace, func(options *metav1.ListOptions) {
 			options.FieldSelector = fields.OneTermEqualSelector("metadata.name", lvm.NodeID).String()
-		}))
+		})
+
+	/*nodeInformerFactory := informers.NewSharedInformerFactoryWithOptions(
+	openebsClient, 0, informers.WithNamespace(lvm.LvmNamespace),
+	informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+		options.FieldSelector = fields.OneTermEqualSelector("metadata.name", lvm.NodeID).String()
+	}))*/
```
Ab-hishek: Can we remove the commented lines?

pkg/mgmt/lvmnode/start.go
```diff
@@ -84,16 +88,18 @@ func Start(controllerMtx *sync.RWMutex, stopCh <-chan struct{}) error {
...
+	/*controller, err := NewNodeControllerBuilder().
+	withKubeClient(kubeClient).
+	withOpenEBSClient(openebsClient).
+	withNodeSynced(nodeInformerFactory).
+	withNodeLister(nodeInformerFactory).
+	withRecorder(kubeClient).
+	withEventHandler(nodeInformerFactory).
+	withPollInterval(60 * time.Second).
+	withOwnerReference(ownerRef).
+	withWorkqueueRateLimiting().Build()*/
```
Ab-hishek: Why are we keeping these commented lines? Any specific reason?

pkg/mgmt/snapshot/builder.go
```diff
@@ -33,15 +33,22 @@ import (
 
 const controllerAgentName = "lvmsnap-controller"
 
+var resource = schema.GroupVersionResource{
+	Group:    "local.openebs.io",
+	Version:  "v1alpha1",
+	Resource: "lvmsnapshots",
```
Ab-hishek: Can we use constants for all these values.

pkg/mgmt/lvmnode/builder.go
```diff
@@ -36,15 +36,23 @@ import (
 
 const controllerAgentName = "lvmnode-controller"
 
+var resource = schema.GroupVersionResource{
+	Group:    "local.openebs.io",
+	Version:  "v1alpha1",
+	Resource: "lvmnodes",
```
Ab-hishek: Use constants for all these fields.

pkg/mgmt/snapshot/snapshot.go
Ab-hishek: We can use the `getStructuredObject` function here.

pkg/mgmt/snapshot/snapshot.go
```diff
@@ -96,23 +104,23 @@ func (c *SnapController) syncSnap(snap *apis.LVMSnapshot) error {
...
+		runtime.HandleError(fmt.Errorf("Couldn't get snaphot object %#v", obj))
 		return
 	}
 
 	if lvm.NodeID != snap.Spec.OwnerNodeID {
 		return
 	}
-	klog.Infof("Got add event for Snap %s/%s", snap.Spec.VolGroup, snap.Name)
+	klog.Infof("Got add event for Snapshot %s/%s", snap.Spec.VolGroup, snap.Name)
+	klog.Infof("lvmsnapshot object to be enqueued by Add handler: %v", snap)
```
Ab-hishek: Was this adding for debugging purpose? If so, we can remove this now.

pkg/mgmt/volume/builder.go
```diff
@@ -33,15 +33,22 @@ import (
 
 const controllerAgentName = "lvmvolume-controller"
 
+var resource = schema.GroupVersionResource{
+	Group:    "local.openebs.io",
+	Version:  "v1alpha1",
+	Resource: "lvmvolumes",
```
Ab-hishek: Same comments as above. Use constants for these.

pkg/mgmt/volume/volume.go
```diff
@@ -49,31 +51,113 @@ func (c *VolController) syncHandler(key string) error {
...
 	if err != nil {
 		return err
 	}
-	VolCopy := Vol.DeepCopy()
+	vol := &apis.LVMVolume{}
+	err = runtimenew.DefaultUnstructuredConverter.FromUnstructured(unstructuredVol.UnstructuredContent(), &vol)
+	//err = runtime.DefaultUnstructuredConverter.FromUnstructured(Vol.UnstructuredContent(), &vol)
+	if err != nil {
+		klog.Infof("err %s, While converting unstructured obj to typed object\n", err.Error())
+	}
```
Ab-hishek: We can use the `getStructuredObject` function here.

pkg/mgmt/volume/volume.go
```diff
@@ -49,31 +51,113 @@ func (c *VolController) syncHandler(key string) error {
...
+		return
+	}
+
+	if lvm.NodeID != newVol.Spec.OwnerNodeID {
+		return
+	}
+
+	if c.isDeletionCandidate(newVol) {
+		klog.Infof("Got update event for deleted Vol %s", newVol.Name)
+		klog.Infof("Deletion timestamp for the volume from UpdateVOl: %v", newVol.ObjectMeta.DeletionTimestamp)
```
Ab-hishek: We can remove this line from the logs.

pkg/mgmt/volume/volume.go
```diff
@@ -49,31 +51,113 @@ func (c *VolController) syncHandler(key string) error {
...
+		c.enqueueVol(newVol)
+	}
+}
+
+// deleteVol is the delete event handler for LVMVolume
+func (c *VolController) deleteVol(obj interface{}) {
+	Vol, ok := c.getStructuredObject(obj)
+	if !ok {
+		unstructuredObj, ok := obj.(*unstructured.Unstructured)
+		if ok {
```
Ab-hishek: We should return some error in case of `!ok`. Right?

pkg/mgmt/lvmnode/lvmnode.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/lvmnode.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/start.go
abhilashshetty04: Resync period is not used to update informer cache after specified interval rather it force reconciles all stored cache. According to the following github issue. Hence 5 minutes seems fine, right?

https://github.com/kubernetes-sigs/controller-runtime/issues/521 

Find this comment:
The problem was that nobody actually understood what this knob did -- it doesn't relist from the API server (which is what most people thought it did). All it does is force a re-reconcile of everything in the cache. Generally, you probably don't want to do this -- it's often expensive, it'll mask subtle bugs, and it's usually not necessary. It's actually probably fine to set this to never, except it might save you in production if you've got bugs in your reconcile logic because you didn't write a level-based controller.

This is the very first comment.

pkg/mgmt/lvmnode/builder.go
abhilashshetty04: I have kept this what was previously there..Not sure what it is. Will check more

pkg/mgmt/lvmnode/start.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/start.go
abhilashshetty04: Incorporated

pkg/mgmt/snapshot/builder.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/builder.go
abhilashshetty04: Incorporated

pkg/mgmt/volume/volume.go
abhilashshetty04: Initially i thought of using it. However, Objects from the queue had to be type asserted to *unstructured.Unstructured datatype. Objects that we get from informer cache using lister are already of the type we are watching. In Above mentioned code unstructuredVol is the object from cache.

Hence have not used that method.

pkg/mgmt/snapshot/snapshot.go
abhilashshetty04: Incorporated. Keeping line 116. Will help track add events in the logs

pkg/mgmt/volume/volume.go
abhilashshetty04: this will help track deleted objects along with timestamp. Seems like a good logger right?

pkg/mgmt/volume/builder.go
abhilashshetty04: Incorporated

pkg/mgmt/snapshot/snapshot.go
abhilashshetty04: Initially i thought of using it. However, Objects from the queue had to be type asserted to *unstructured.Unstructured datatype. Objects that we get from informer cache using lister are already of the type we are watching. In Above mentioned code unstructuredVol is the object from cache.

Hence have not used that method.

pkg/mgmt/snapshot/snapshot.go
abhilashshetty04: Noted. Will be keeping line 116. Seems like a needed logger.

pkg/mgmt/volume/volume.go
abhilashshetty04: Will add them in one logger.

pkg/mgmt/volume/volume.go
abhilashshetty04: Will look into it. I think logger is not required for it though..right?

abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/builder.go
```diff
@@ -17,34 +17,47 @@
...
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
+	"time"
 )
 
 const controllerAgentName = "lvmnode-controller"
 
+const (
```
niladrih, line 37: We may have all of the constants in the same `const` block.

pkg/mgmt/lvmnode/lvmnode.go
niladrih: @abhilashshetty04 you can use Right click --> `Go Tools` --> Goimports File to fix the import order in your IDE (Goland).

pkg/mgmt/lvmnode/lvmnode.go
```diff
@@ -54,13 +56,18 @@ func (c *NodeController) syncHandler(key string) error {
 // LVMNode
 func (c *NodeController) syncNode(namespace string, name string) error {
 	// Get the node resource with this namespace/name
-	cachedNode, err := c.NodeLister.LVMNodes(namespace).Get(name)
+	cachedNode, err := c.NodeLister.Namespace(namespace).Get(name)
```
niladrih, line 61: Can we name this variable more appropriately?

pkg/mgmt/snapshot/snapshot.go
```diff
@@ -149,10 +162,26 @@ func (c *SnapController) deleteSnap(obj interface{}) {
...
+	if ok {
+		snap := &apis.LVMSnapshot{}
+		err := runtimenew.DefaultUnstructuredConverter.FromUnstructured(unstructuredInterface.UnstructuredContent(), &snap)
+		if err != nil {
+			fmt.Printf("err %s, While converting unstructured obj to typed object\n", err.Error())
+			return nil, false
+		}
+		return snap, true
+	}
+	runtime.HandleError(fmt.Errorf("couldnt type assert obj: %#v to unstructured obj", obj))
```
niladrih: Use klog instead of fmt

pkg/mgmt/lvmnode/start.go
```diff
@@ -84,16 +82,7 @@ func Start(controllerMtx *sync.RWMutex, stopCh <-chan struct{}) error {
...
-		withKubeClient(kubeClient).
-		withOpenEBSClient(openebsClient).
-		withNodeSynced(nodeInformerFactory).
-		withNodeLister(nodeInformerFactory).
-		withRecorder(kubeClient).
-		withEventHandler(nodeInformerFactory).
-		withPollInterval(60 * time.Second).
-		withOwnerReference(ownerRef).
-		withWorkqueueRateLimiting().Build()
+	nodeCtrller := newNodeController(kubeClient, openebsClientNew, nodeInformerFactory, ownerRef)
```
niladrih: Why are we changing the variable name. The other controllers are named `controller`.
Following consistent semantics will make it easier to read and use tools like `grep`.

pkg/mgmt/snapshot/snapshot.go
```diff
@@ -124,23 +130,30 @@ func (c *SnapController) updateSnap(oldObj, newObj interface{}) {
...
-		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
-		if !ok {
-			runtime.HandleError(fmt.Errorf("Couldn't get object from tombstone %#v", obj))
-			return
-		}
-		snap, ok = tombstone.Obj.(*apis.LVMSnapshot)
-		if !ok {
-			runtime.HandleError(fmt.Errorf("Tombstone contained object that is not a lvmsnap %#v", obj))
+		unstructuredObj, ok := obj.(*unstructured.Unstructured)
+		if ok {
```
niladrih: Using if-else is not the recommended way...
Try using a pattern like so...
```go
unstructuredObj, ok := obj.(*unstructured.Unstructured)

if !ok {
        runtime.HandleError(fmt.Errorf("couldnt type assert obj: %#v to unstructured obj", obj))
	return
}

tombStone := cache.DeletedFinalStateUnknown{}
err := runtimenew.DefaultUnstructuredConverter.FromUnstructured(unstructuredObj.UnstructuredContent(), &tombStone)
if err != nil {
	runtime.HandleError(fmt.Errorf("couldn't get object from tombstone %#v", obj))
	return
}
snap, ok = tombStone.Obj.(*apis.LVMSnapshot)
if !ok {
	runtime.HandleError(fmt.Errorf("tombstone contained object that is not a lvmvolume %#v", obj))
	return
}
```

pkg/mgmt/volume/volume.go
```diff
@@ -49,31 +51,116 @@ func (c *VolController) syncHandler(key string) error {
...
+			if err != nil {
+				runtime.HandleError(fmt.Errorf("couldn't get object from tombstone %#v", obj))
+				return
+			}
+			Vol, ok = tombStone.Obj.(*apis.LVMVolume)
+			if !ok {
+				runtime.HandleError(fmt.Errorf("tombstone contained object that is not a lvmvolume %#v", obj))
+				return
+			}
+		} else {
```
niladrih: if-else is not good practice. Read comment at a previous point with similar pattern.

pkg/mgmt/volume/volume.go
```diff
@@ -49,31 +51,116 @@ func (c *VolController) syncHandler(key string) error {
...
 		runtime.HandleError(err)
 		return
 	}
 	c.workqueue.Add(key)
+}
 
+//Obj from queue is not readily in lvmvol type. This function would convert obj into lvmvolume type.
+func (c *VolController) getStructuredObject(obj interface{}) (*apis.LVMVolume, bool) {
+	unstructuredInterface, ok := obj.(*unstructured.Unstructured)
+	if ok {
```
niladrih: Default case should not be error handling.
Better to use...
```go
if !ok {
        // Handle error
}

// Code for when type assertion succeeds
```

niladrih (changes requested):
Few comments @abhilashshetty04.

pkg/mgmt/lvmnode/lvmnode.go
abhilashshetty04: This is the function dynamic lister exposes. This cannot be changed.

pkg/mgmt/lvmnode/start.go
abhilashshetty04: renamed variables for other controllers appropriately .

pkg/mgmt/lvmnode/lvmnode.go
niladrih: **e.g.**

cachedNode --> cachedObj

pkg/mgmt/snapshot/snapshot.go
abhilashshetty04: In this case it is not feasible to use klog as klog doesnt return error object. We need to pass error object to HandleError. fmt.Errorf will return err object.

pkg/mgmt/volume/volume.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/builder.go
niladrih: Let's remove the builder. 👍

pkg/mgmt/snapshot/snapshot.go
niladrih: Sorry, I misread. Could you use github.com/pkg/errors package, as we are doing elsewhere here. Basically 
```go
errors.Errorf("couldnt type assert obj: %#v to unstructured obj", obj)
```

niladrih: Using different loggers will give rise inconsistency in the logging order. All log messages must be in the same logging queue.

pkg/mgmt/lvmnode/builder.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/lvmnode.go
niladrih: Name seems appropriate as it is -- cachedNode ✔️ .

pkg/mgmt/snapshot/snapshot.go
abhilashshetty04: Incorporated.

pkg/mgmt/lvmnode/start.go
abhilashshetty04: Using controller for all controller variables.

pkg/mgmt/snapshot/snapshot.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/builder.go
Ab-hishek: @abhilashshetty04 You can use this setting in your Goland IDE -> https://www.jetbrains.com/help/go/creating-and-optimizing-imports.html#use-the-goimports-local-grouping

pkg/mgmt/lvmnode/builder.go
```diff
@@ -17,34 +17,46 @@
...
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
+	"time"
+)
+
+const (
+	controllerAgentName = "lvmnode-controller"
+	GroupOpenebsIO      = "local.openebs.io"
+	VersionV1alpha1     = "v1alpha1"
+	Resource            = "lvmnodes"
```
Ab-hishek, line 41: Can `Resource` be renamed to `lvmNodeResource`?

pkg/mgmt/lvmnode/builder.go
```diff
@@ -17,34 +17,46 @@
...
 )
 
-const controllerAgentName = "lvmnode-controller"
+var resource = schema.GroupVersionResource{
```
Ab-hishek: Instead of just naming it as resource, can we rename it to `lnResource` where `ln` would represent `lvm node`. I have seen this kind of naming standard being followed in a lot of places for dynamic resource controllers.

pkg/mgmt/lvmnode/builder.go
abhilashshetty04: Incorporated

pkg/mgmt/lvmnode/lvmnode.go
abhilashshetty04: Incorporated

Ab-hishek (approved):
lgtm. Thanks @abhilashshetty04

pkg/mgmt/lvmnode/lvmnode.go
```diff
@@ -43,7 +47,7 @@ func (c *NodeController) syncHandler(key string) error {
 	// Convert the namespace/name string into a distinct namespace and name
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
-		runtime.HandleError(fmt.Errorf("invalid resource key: %s", key))
+		runtime.HandleError(errors.Errorf("invalid resource key: %s", key))
```
niladrih: errors --> fmt

niladrih (changes requested):
One little nitpick

pkg/mgmt/snapshot/snapshot.go
niladrih: errors is not the default errors package for these modules, this change is not necessary.

niladrih (approved):

Issue #194, opened 2022-06-14
Modify ratelimiter for controlled exponential backoff in lvmvolume controller


abhilashshetty04:
With the default ratelimiter DefaultControllerRateLimiter()  being used while creating controller for lvmvolumes. It re queues items after 5ms. Which means reconciliation happens after ~5ms. Which is unnecessary.

We need to look for ratelimiter which can increase the period between each retries.

abhilashshetty04:
Closing issue as Pr https://github.com/openebs/lvm-localpv/pull/195 is merged

PR #195, opened 2022-06-14
fix: Modify ratelimiter for controlled exponential backoff in lvmvolume controller

abhilashshetty04:
We have changed the rate limiter which was used in snapshot controller. This rate limiter re queues failed items after 5 secs for first 12 attempts. Then objects are re queued after 30 secs.

This is the new rate limiter being used:
`rateLimiter := workqueue.NewItemFastSlowRateLimiter(5*time.Second, 30*time.Second, 12)`

We used to use DefaultControllerRateLimiter() before this which would re queue failed items after 5ms delay which is really not required for Lvm localpv system. reconciliation failures are mostly misconfigurations on worker nodes where vol/snapshot being created doesn't have any network calls as far as node plugin implementation is concerned.

**Checklist:**
- [X] Fixes [#<194>](https://github.com/openebs/lvm-localpv/issues/194)
- [X] PR Title follows the convention of  `<type>(<scope>): <subject>`

### DIFF
--- a/pkg/mgmt/volume/builder.go
+++ b/pkg/mgmt/volume/builder.go
@@ -17,6 +17,8 @@ limitations under the License.
 package volume
 
 import (
+	"time"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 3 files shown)


pkg/mgmt/volume/builder.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package volume
 
 import (
+	"time"
```
pawanpraka1, line 20: why is this needed?

Ab-hishek: time package is used below @pawanpraka1 for defining the rate limits.

pawanpraka1 (approved):
looks good.

pawanpraka1:
@abhilashshetty04 can you DCO signoff this PR?

Ab-hishek (approved):
LGTM

niladrih (approved):

niladrih (changes requested):
DCO sign-off is required

niladrih (approved):

abhilashshetty04:
@niladrih Commit is DCO signed off

PR #229, opened 2023-05-15
fix(localpv): expose Kubernetes client QPS and Burst as flags

barkbay:

While running some performance tests, I noticed that there seems to be a bottleneck that is significantly delaying volumes creation. Here are the results when trying to create 5 volumes per second (see the test scenario at the end of this comment):

image

After some investigation, I noticed that the bottleneck was the Kubernetes client throttling requests.

**What this PR does?**:

This PR allows the user to set the QPS and Burst value for the K8S client. Running the same performance test as above with a `qps` of `50` and a `burst` of `100` leads to the following result:

image

No, if no `--kube-api-qps` or `--kube-api-burst` are provided, then the behavior is exactly the same as today.

Here is [ClusterLoader2](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2#clusterloader) scenario used for the performance test presented above:

```yaml
## Default expected number of PVC is 300
{{$PVCS := DefaultParam .PVCS 300}}

## Set a default timeout of 30 minutes
{{$TIMEOUT := DefaultParam .TIMEOUT "1800s"}}

name: openebs-stress

namespace:
  number: 1

tuningSets:
  - name: UniformQPS
    qpsLoad:
      qps: 5

steps:
  - name: Create PVCs
    phases:
      - namespaceRange:
          min: 1
          max: 1
        replicasPerNamespace: {{$PVCS}}
        tuningSet: UniformQPS
        objectBundle:
          - basename: volume-test
            objectTemplatePath: "pvc.yaml"
  - name: Waiting for PVs to be bound
    measurements:
      - Identifier: WaitForPVCsToBeBound
        Method: WaitForBoundPVCs
        Params:
          desiredPVCCount: {{$PVCS}}
          apiVersion: v1
          labelSelector: group = volume-test
          timeout: {{$TIMEOUT}}
```

With `pvc.yaml`:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: {{.Name}}
  labels:
    group: volume-test
spec:
  storageClassName: "openebs-immediate" ## Update to reflect you cs
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
```

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -97,6 +97,14 @@ func main() {
 		&config.DisableExporterMetrics, "disable-exporter-metrics", true, "Exclude metrics about the exporter itself (process_*, go_*).",
 	)
 
+	cmd.PersistentFlags().IntVar(
+		&config.KubeAPIQPS, "kube-api-qps", 0, "QPS to use while talking with Kubernetes API server.",
+	)
+
+	cmd.PersistentFlags().IntVar(
+		&config.KubeAPIBurst, "kube-api-burst", 0, "Burst to allow while talking with Kubernetes API server.",
+	)
+
 	config.RIopsLimitPerGB = cmd.PersistentFlags().StringSlice(
 		"riops-per-gb", []string{},
 		"Read IOPS per GB limit to use for each volume group prefix, "+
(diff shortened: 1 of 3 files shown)


niladrih:
Thanks you for your contribution! This change has been taken up as a part of PR #234, closing this PR.

PR #256, opened 2023-09-03
feat(NodeController): made node polling interval configurable

### LINKED ISSUE #255: Allow configurable polling interval in NodeController
Current polling interval in NodeController is hardcoded to be 60 seconds. This means that the latency for node updates, and in turn CSIStorageCapacity updates, is up to 60s, even after a PVC was bound. As a result, custom k8s schedulers that are aware of storage can make bad decisions - either unnecessary evictions, or impossible allocations.

Make pollInterval in `lvmnode.NodeController` configurable, with the default being 60s like today, and allow users to customize according to their use-case.

**Environment:**
- LVM Driver version: 1.2.0
- Kubernetes version (use `kubectl version`): v1.27.4

itsomri:
This PR makes the polling interval of the node configurable, allowing users to get a higher update rate for their csistoragecapacities, and preventing scheduling issue.

**What this PR does?**:
This PR adds a new flag, `--node-polling-interval`, with a default value of 60s (same as the currant hard-coded value), which controls the polling rate of the node.

No.

1. Verified that the default remains 60s, when not using the flag.
2. Verified that changing the value of the flag changes the polling rate.

N/A

**Checklist:**
- [X] Fixes #255
- [X] PR Title follows the convention of  `<type>(<scope>): <subject>`

### DIFF
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -66,7 +66,7 @@ func NewNode(d *CSIDriver) csi.NodeServer {
 
 	// start the lvm node resource watcher
 	go func() {
-		err := lvmnode.Start(d.config.NodeControllerPollingInterval, &ControllerMutex, stopCh)
+		err := lvmnode.Start(&ControllerMutex, stopCh, d.config.NodeControllerPollingInterval)
 		if err != nil {
 			klog.Fatalf("Failed to start LVM node controller: %s", err.Error())
 		}
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -66,7 +66,7 @@ func NewNode(d *CSIDriver) csi.NodeServer {
 
 	// start the lvm node resource watcher
 	go func() {
-		err := lvmnode.Start(&ControllerMutex, stopCh)
+		err := lvmnode.Start(d.config.NodeControllerPollingInterval, &ControllerMutex, stopCh)
 		if err != nil {
 			klog.Fatalf("Failed to start LVM node controller: %s", err.Error())
 		}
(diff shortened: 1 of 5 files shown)


abhilashshetty04:
Hi @omric-runai , Thanks for the PR.  lgtm

abhilashshetty04 (approved):

Abhinandan-Purkait (commented):
@avishnu @abhilashshetty04 Since this is configurable, I think we need a helm value and template change to accommodate this?

pkg/driver/agent.go
```diff
@@ -66,7 +66,7 @@ func NewNode(d *CSIDriver) csi.NodeServer {
 
 	// start the lvm node resource watcher
 	go func() {
-		err := lvmnode.Start(&ControllerMutex, stopCh)
+		err := lvmnode.Start(d.config.NodeControllerPollingInterval, &ControllerMutex, stopCh)
```
avishnu: Can you add the new argument as the last argument instead of the first?

avishnu:
> @avishnu @abhilashshetty04 Since this is configurable, I think we need a helm value and template change to accommodate this?

Yes, we need a configurable parameter in the helm values.yaml

pkg/driver/agent.go
itsomri: Fixed

Abhinandan-Purkait:
@omric-runai Thanks for the PR. Would you be interested in raising a helm chart change for allowing configuring this from Helm Chart? You can raise a separate PR for that if you want.

Abhinandan-Purkait (approved):

PR #260, opened 2023-10-08
Changed the logic for comparing volsize and updatesize when scaling up.

### LINKED ISSUE #220: [PVC Expand]A special case will case pvc updated, but not expand lv by lvm-node
## Description
As https://github.com/openebs/lvm-localpv/blob/v0.8.x/design/lvm/images/resize_sequence_diagram.jpg show
If happen:
1. step "Mark PVC As Filesystem resize required" Failed.
2. Skip NodeExpandVolume
3. run whole steps again

But when executing this process for the second time, it will happen:
  1. ControllerExpandVolume return fsResizeRequired=nil because of https://github.com/openebs/lvm-localpv/blob/develop/pkg/driver/controller.go#L544
  4. step "Mark PVC As Filesystem resize required" skipped, and run markPVCResizeFinished.(mark pvc update finished)
  5. skip NodeExpandVolume

## Conclusion 
So In above case, it will happen **pvc updated finsihed but lv didn't resize**

## Describe
lv can resize normally even if above case happen.

esimest:
this should fix #220

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -553,20 +553,17 @@ func (cs *controller) ControllerExpandVolume(
 	 * to the specified volume ID is already larger than or equal to the target
 	 * capacity of the expansion request, the plugin should reply 0 OK.
 	 */
-	if volsize >= updatedSize {
-		return csipayload.NewControllerExpandVolumeResponseBuilder().
-			WithCapacityBytes(volsize).
-			Build(), nil
+	if volsize < updatedSize {
+		if err := lvm.ResizeVolume(vol, updatedSize); err != nil {
+			return nil, status.Errorf(
+		    	codes.Internal,
+		    	"failed to handle ControllerExpandVolumeRequest for %s, {%s}",
+		    	volumeID,
+		    	err.Error(),
+		    )
+		}
 	}
-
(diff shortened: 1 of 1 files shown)


pkg/driver/controller.go
```diff
@@ -553,20 +553,17 @@ func (cs *controller) ControllerExpandVolume(
...
-		)
+	if volsize < updatedSize {
+		if err := lvm.ResizeVolume(vol, updatedSize); err != nil {
+			return nil, status.Errorf(
+		    	codes.Internal,
+		    	"failed to handle ControllerExpandVolumeRequest for %s, {%s}",
+		    	volumeID,
+		    	err.Error(),
+		    )
+		}
```
Abhinandan-Purkait, line 564: I am not seeing a reason to have this change. Also it removes the bit which makes the call idempotent. I would request you to explain this in the issue.

Abhinandan-Purkait: And this condition `volsize < updatedSize` already gets handled by the sidecars, we don't need to do it again.

Abhinandan-Purkait (commented):
How is it different from what we have already? I understand it's a refactor. But can you please explain what was the behaviour without this?

Abhinandan-Purkait:
Closing this since we did not receive any response. Please feel free to reopen this.

PR #280, opened 2024-01-10
feat(charts): disable hostNetwork on node plugin

### LINKED ISSUE #274: Allow disabling hostNetwork: true
Reduce amount of open ports on k8s nodes.
As far as I understood the only port is exposed is 9500 and used only for metrics.
So it seems it's possible to collect it using pod or service scrapper. 

Helm value that would allow to disable `hostNetwork: false`

For such case it would also useful to allow modify pod and service annotations throw chart values

**Environment:**
- LVM Driver version

Abhinandan-Purkait:
Resolves #274 

This allows hostNetwork on lvm-csi-node to be configurable, with default as false.

### DIFF
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: lvm-localpv
 description: CSI Driver for dynamic provisioning of LVM Persistent Local Volumes.
-version: 1.4.0
+version: 1.4.1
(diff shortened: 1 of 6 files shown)


abhilashshetty04 (approved):

avishnu (approved):

Abhinandan-Purkait:
https://github.com/openebs/lvm-localpv/pull/284 Cherry-picked

PR #289, opened 2024-03-27
fix: crds getting removed on helm upgrades

Abhinandan-Purkait:
#### Why do we need this PR?

It was seen that on upgrade to 1.5.x the CRDs get removed causing the removal of CRs. This would prevent it.
It also adds annotations to let helm keep the CRDs on uninstall, which is configurable.

### DIFF
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: lvm-localpv
 description: CSI Driver for dynamic provisioning of LVM Persistent Local Volumes.
-version: 1.5.0
+version: 1.5.1
 appVersion: 1.5.0
 icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/openebs/icon/color/openebs-icon-color.png
 home: https://openebs.io/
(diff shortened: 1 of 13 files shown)


avishnu (approved):
lgtm

deploy/helm/charts/charts/crds/templates/_helpers.tpl
```diff
@@ -1,17 +1,18 @@
+{{/* vim: set filetype=mustache: */}}
```
tiagolobocastro, line 1: this is not a mustache file is it?

tiagolobocastro (approved):

deploy/helm/charts/charts/crds/templates/_helpers.tpl
Abhinandan-Purkait: It's somewhat similar to mustache grammar. I got this from mayastor extensions, tbh. 
ref:- https://github.com/openebs/mayastor-extensions/blob/4ab5529064e4623dcbf5510a0795043c25cc5979/chart/charts/crds/templates/_helpers.tpl#L1

deploy/helm/charts/Chart.yaml
```diff
@@ -16,11 +16,5 @@ sources:
...
   - name: crds
-    version: "1.5.0"
-maintainers:
-  - name: prateekpandey14
-    email: prateek.pandey@mayadata.io
-  - name: pawanpraka1
-    email: pawan@mayadata.io
-  - name: iyashu
-    email: yashpal.c1995@gmail.com
+    version: "1.5.1"
```
niladrih: nit: this shouldn't have the double quotes because we're typically not using quotes in chart versions.

niladrih (commented):
Left a comment.

Issue #93, opened 2021-06-10
Enable monitoring on thin pool to allow auto-extend threshold automatically 


prateekpandey14:
**Describe the problem/challenge you have**
- Automatic extension requires **dmeventd** (device mapper event daemon) service to be running to continuously monitor the LVs
But in case of container based LVM is not able to communicate with dmeventd service, the monitoring gets disable, and auto extend of thin pools did not work as expected.
```  
 $ lvs -o+seg_monitor 
  LV               VG       Attr           LSize   Pool   ....          Monitor    
  lvmvg_thinpool  lvmvg  twi-aotz--        2.07g          ....          not monitored   
```
-  For a workaround currently its requires monitoring to be explicitly enabled once from the host node, on per thin pool bases. 
```sh
$ lvchange --monitor y vg-name/lvmvg_thinpool
```
-  Without this step, automatic extension of the logical thin pool will not occur, regardless of any settings in the `lvm.conf`.

**Describe the solution you'd like**
We need to find a way to communicate to the `dmeventd` service from `lvm-csi-plugin` container , which will help to automatically enable the monitoring and no manual action required.

ianroberts:
> * For a workaround currently its requires monitoring to be explicitly enabled once from the host node, on per thin pool bases.
> 
> ```shell
> $ lvchange --monitor y vg-name/lvmvg_thinpool
> ```

... and this command can only be run _after_ the first LVM local PV has been provisioned on that particular node.  If I try to enable monitoring on a node where the thin pool exists but has not yet had any local PVs provisioned on it, I get

```
# lvchange --monitor y localhd/localhd_thinpool
  Logical volume localhd/localhd_thinpool is not active.
```

andrey-podko:
@ianroberts Funny fact: Piraeus/linbit has same behavior - first creation of thin volume on thinpool disables "monitor" option

dsharma-dc:
This has been looked at recently while starting to add some new CI test around thin provisioning. I'd think that a good reason to not automatically enable monitoring on thinpool is that some users may not want to have thinpool grow automatically or grow beyond a certain limit. There is no config in lvm.conf to define a cap/limit of auto-extend. e.g. on a 10G VG, they may not want a 5G thinpool to grow beyond 7G. I don't see a way to restrict that. Hence, users can manually enable monitoring depending on the use case and required behaviour of thinpool.

ianroberts:
[...]

PR #309, opened 2024-06-12
test creating sized snapshots for filesystem and block volumes

dsharma-dc:
## Pull Request template

Exercises the snapSize behaviour of volume snapshot class.

**What this PR does?**:
Tests creating sized snapshots for filesystem and raw block volumes.

No

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/tests/run_test.go
+++ b/tests/run_test.go
@@ -37,6 +37,18 @@ spec:
 `
 )
 
+const (
+	sizedsnapYAML = `apiVersion: snapshot.storage.k8s.io/v1
+kind: VolumeSnapshot
+metadata:
+  name: #snapname
+spec:
+  volumeSnapshotClassName: lvmpv-snapclass-sized
+  source:
+    persistentVolumeClaimName: #pvcname
+`
+)
+
 func execAtLocal(cmd string, input []byte, args ...string) ([]byte, []byte, error) {
 	var stdout, stderr bytes.Buffer
 	command := exec.Command(cmd, args...)
@@ -68,17 +80,17 @@ func verifySnapshotCreated(snapName string) bool {
 	return true
 }
 
-func createSnapshot(pvcName, snapName string) {
+func createSnapshot(pvcName, snapName string, snapYaml string) {
 	By("creating snapshot for a pvc " + pvcName)
 
-	tyaml := strings.Replace(snapYAML, "#pvcname", pvcName, -1)
+	tyaml := strings.Replace(snapYaml, "#pvcname", pvcName, -1)
 	yaml := strings.Replace(tyaml, "#snapname", snapName, -1)
 
 	stdout, stderr, err := kubectlWithInput([]byte(yaml), "apply", "-n", OpenEBSNamespace, "-f", "-")
 	Expect(err).ShouldNot(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
 }
 
-func deleteSnapshot(pvcName, snapName string) {
+func deleteSnapshot(pvcName, snapName string, snapYaml string) {
 	By("deleting the snapshot " + snapName)
 
 	tyaml := strings.Replace(snapYAML, "#pvcname", pvcName, -1)
--- a/deploy/sample/lvmsnapclass.yaml
+++ b/deploy/sample/lvmsnapclass.yaml
@@ -6,3 +6,14 @@ metadata:
     snapshot.storage.kubernetes.io/is-default-class: "true"
 driver: local.csi.openebs.io
 deletionPolicy: Delete
+---
+kind: VolumeSnapshotClass
+apiVersion: snapshot.storage.k8s.io/v1
+metadata:
+  name: lvmpv-snapclass-sized
+  annotations:
+    snapshot.storage.kubernetes.io/is-default-class: "false"
+driver: local.csi.openebs.io
+deletionPolicy: Delete
+parameters:
+  snapSize: 80%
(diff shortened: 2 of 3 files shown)


dsharma-dc:
This PR is based on top of branch used for PR 308 and hence we see that commit here as well. That commit doesn't need review against this PR.

pkg/lvm/lvm_util.go
```diff
@@ -53,13 +53,15 @@ const (
...
 	LVExtend = "lvextend"
+	LVChange = "lvchange"
 	LVList   = "lvs"
 
 	PVList = "pvs"
 	PVScan = "pvscan"
 
 	YES        = "yes"
 	LVThinPool = "thin-pool"
+	SUDO       = "sudo"
```
tiagolobocastro: Was sudo not required previously?

pkg/lvm/lvm_util.go
```diff
@@ -868,13 +870,46 @@ func decodeLvsJSON(raw []byte) ([]LogicalVolume, error) {
...
+	if err != nil {
+		klog.Errorf("lvm: error while running command %s %v: %v", LVChange, args, err)
+		return err
+	}
+
+	return nil
+}
+
+// Get all, or a particular LVM LV. If lvname is nil, all LVs are listed, otherwise the requested one.
+func ListLVMLogicalVolume(lvname string) ([]LogicalVolume, error) {
```
tiagolobocastro: I'd suggest using different functions rather than overload and having to pass empty string as you have above

pkg/lvm/lvm_util.go
```diff
@@ -891,12 +926,20 @@ func ListLVMPhysicalVolume() ([]PhysicalVolume, error) {
...
 
+	cmd := PVList
 	args := []string{
 		"--options", "pv_all,vg_name",
 		"--reportformat", "json",
 		"--units", "b",
 	}
-	output, _, err := RunCommandSplit(PVList, args...)
+
+	// Are we running privileged or not?
```
tiagolobocastro: this is duplicated, add function to do this?

ci/ci-test.sh
```diff
@@ -97,6 +97,10 @@ sudo vgcreate foreign_lvmvg "${foreign_disk}" --config="${FOREIGN_LVM_CONFIG}"
 sudo modprobe dm-snapshot
 sudo modprobe dm_thin_pool
 
+# Set the configuration for thin pool autoextend in lvm.conf
+sudo sed -i '/^[^#]*thin_pool_autoextend_threshold/ s/= .*/= 50/' /etc/lvm/lvm.conf
```
tiagolobocastro: Shouldn't this be part of the particular test itself?

tests/utils.go
```diff
@@ -706,3 +707,38 @@ func createNodeDaemonSet(ds *appsv1.DaemonSet) {
...
+	gomega.Expect(err).To(gomega.BeNil(), "list LVM LV")
+	gomega.Expect(len(vol)).To(gomega.Equal(1), "retrieve LVM LV")
+
+	thinpool := VOLGROUP + "/" + vol[0].PoolName
+
+	args := []string{
+		"--monitor", "y",
+	}
+
+	err = lvm.ChangeLVMLogicalVolumeAttrs(thinpool, args)
```
tiagolobocastro: shouldn't this be enabled by the product itself?

pkg/lvm/lvm_util.go
dsharma-dc: It usually runs from privileged container so `sudo` isn't required.  I added that so the ci script can also run this. However, I've reworked this a bit and there are no more changes required in `lvm_util.go` . Updated diff in PR 308. The commit ([test: thinpool auto-extend test](https://github.com/openebs/lvm-localpv/pull/309/commits/2ca22d9dcbe1384119af5fca96c07b007742f9c3)) in this PR is from PR 308 since this PR's branch is on top of that.

pkg/lvm/lvm_util.go
dsharma-dc: already updated diff in PR 308.

ci/ci-test.sh
dsharma-dc: We just need this configured once, and it's simpler to do this as part of test setup, just like in reality users have to manually pre-configure this.

tests/utils.go
dsharma-dc: Enabling by default in product is a tricky decision depending on who wants it and who doesn't. Some discussion here: https://github.com/openebs/lvm-localpv/issues/93

pkg/lvm/lvm_util.go
dsharma-dc: dropped this from diff anyway

tests/utils.go
tiagolobocastro: I didn't mean by default right. I meant that product should add by itself, depending on user configuration.
But from this I take it that today the product would never do that, and users have to manually enable it?
Then it's fine to test it as such.

dsharma-dc: > But from this I take it that today the product would never do that

That's right, currently we have no provision to do that. There are a couple of enhancements identified in this area.

tests/run_test.go
```diff
@@ -68,20 +80,31 @@ func verifySnapshotCreated(snapName string) bool {
 	return true
 }
 
-func createSnapshot(pvcName, snapName string) {
+func createSnapshot(pvcName, snapName string, sized bool) {
```
tiagolobocastro: This size parameter was a bit confusing to figure out. Suggest you simply pass the const yaml var or the classname.

tiagolobocastro (approved):

tests/run_test.go
dsharma-dc: > This size parameter was a bit confusing to figure out. Suggest you simply pass the const yaml var or the classname.

Passing the yaml name var now.

PR #313, opened 2024-06-14
test(volgroup_vgpattern_bdd): adding feature file

abhilashshetty04:
**What this PR does?**:
This PR includes bdd specification for volgroup and vgpattern based test

NA
NA

### DIFF
--- a/tests/bdd/volgroup_vgpattern_bdd.feature
+++ b/tests/bdd/volgroup_vgpattern_bdd.feature
@@ -1,0 +1,179 @@
+Feature: Validate vgpattern and volgroup based scheduling.
+  Background:
+      Given Single lvmnode cluster is configured
+      And lvm vg "lvmvg" is created on a node
+
+      Scenario: test sc with volgroup present on the node
+          Given a sc is created with volgroup as "lvmvg"
+          And pvc is created referencing the same sc
+          Then lvmvolume cr should get created
+          And lvmvolume cr state should be Ready
+
+      Scenario: test sc with volgroup not present on the node
+          Given a sc is created with volgroup as "lvmvgn"
+          And pvc is created referencing the same sc
+          Then pvc should be in Pending
+          And lvmvolume cr should not get created
+
+  
+  Background:
+      Given Single lvmnode cluster is configured
+      And lvm vg "lvmvg" and "lvmvg1" is created on a node
+
+      Scenario: test sc with vgpattern matching atleast one vg
+          Given a sc is created with vgpattern "lvmvg1.*"
+          When a pvc is created by referencing the sc
+          Then pvc should be Bound
+          And lvmvolume cr should be in "lvmvg1" vg
+
+  
+      Scenario: test sc with vgpattern matching no vg
+          Given a sc is created with vgpattern "lvmvgn.*"
+          When a pvc is created by referencing the sc
+          Then pvc should not be Bound
+          And lvmvolume cr should not be created
+
+   Background:
+        Given Single lvmnode cluster is configured
+        And both <vg> are created with their respective <size>
+
+        Scenario: test pvc with larger size then vg free space
+            Given sc is created with volgroup parameter as "lvmvg"
+            When a pvc is created with size "1.3G"
+            Then pvc should not be bound
+            And lvmvolume cr should not be created
+            When "lvmvg" is extended to "1.5G"
+            Then pvc should be Bound
+            And lvmvolume cr should have expected vg value
+
+            Examples:
+                |  vg        | size     |
+                | lvmvg      | 1G       |
+                | lvmvg1     | 2G       |
+
+  
+  Background:
+      Given a 3 lvmnode cluster is configured
+      And each node <node> has <vg> created with same size 
+
+      Scenario: test sc with volgroup present only on one node
+          Given a sc is created with volgroup as "lvmvg1" with no topology
+          When pvc is created referencing the same sc
+          Then pvc should be in Pending state
+          And lvmvolume cr should not be created
+
+      Scenario: test sc with volgroup present on one node with allowed topology
+          Given a sc is created with volgroup parameter as "lvmvg1"
+          And sc has topology key with "node1" value
+          When pvc is created referencing the same sc
+          Then pvc should be bound 
+          And lvmvolume cr should be Ready
+          And lvmvolume cr should have expected owner and vg value
+
+       Scenario: test sc with volgroup not present on any node
+          Given a sc is created with volgroup parameter as "lvmvgn"
+          When pvc is created referencing the same sc
+          Then pvc should be in Pending state
+          And lvmvolume cr should not be created
+
+        Examples:
+            |  node        | vg           |
+            | node1        | lvmvg1       |
+            | node2        | lvmvg2       |
+            | node3        | lvmvg3       |
+       
+
+  Background:
+      Given a 3 lvmnode cluster is configured
+      And each node <node> has <vg> created with <size>
+
+      Scenario: test sc with volgroup present on 2 nodes with varying free space and allowed topology
+          Given a sc is created with volgroup parameter as "lvmvg"
+          And node1 and node2 is in allowed topology
+          When a pvc is created with "1.3G" size
+          Then pvc should be Bound
+          And lvmvolume cr should be in Ready state
+          And lvmvolume cr should have expected owner and vg value
+
+      Examples:
+        |  node        | vg          |  size    |
+        | node1        | lvmvg       |  1G      |
+        | node2        | lvmvg       |  1.5G    |
+        | node3        | lvmvgdiff   |  1G      |
+
+        
+      Scenario: test sc with volgroup present on 2 nodes with less space and allowed topology
+          Given a sc is created with volgroup parameter as "lvmvg"
+          And node1 and node2 is in allowed topology
+          When a pvc is created with "1.3G" size
+          Then pvc should be Pending
+          And lvmvolume cr should not be created
+          When node2 vg is extended to <1.5G>
+          Then pvc should be Bound
+          And lvmvolume cr should be created
+          And lvmvolume cr should have expected owner and vg value
+
+      Examples:
+        |  node        | vg          |  size    |
+        | node1        | lvmvg       |  1G      |
+        | node2        | lvmvg       |  1G      |
+        | node3        | lvmvgdiff   |  1G      |
+
+  Background:
+
+      Given a 3 lvmnode cluster is configured
+      And each lvm node <node> has <vg> and <vg1> with <size> and <size1> respectively 
+
+      Scenario: test sc with volgroup present on 2 nodes with varying free space and allowed topology
+          Given a sc is created with volgroup parameter as “lvmvg“
+          And using WaitForFirstConsumer as binding mode
+          And pvc is created with size “1.5G“
+          Then lvmvolume cr should not be created
+          When an app is created referencing the pvc
+          Then pvc should be created
+          And App should be running
+          And lvmvolume cr should be Ready
+          And lvmvolume cr should have expected owner and vg value
+
+       Examples:
+            |  node      | vg             |  size      |  vg1       |   size1     |
+            | node1      | lvmvg          |  1G        | lvmvg1     |     2G      |
+            | node2      | lvmvg          |  1.6G      | lvmvg1     |     2G      |
+            | node3      | lvmvgdiff      |  1G        | lvmvg1     |     2G      |
+      
+  Background:
+
+      Given a 3 lvmnode cluster is configured
+      And each node <node> has <vg> and <vg2> created with same size
+
+      Scenario: test sc with vgpattern matching vg present in all node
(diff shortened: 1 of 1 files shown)


tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,131 @@
...
+Feature: Validate vgpattern and volgroup based scheduling.
+  Background:
+      Given Single lvmnode cluster is configured
+      And lvm vg "lvmvg" is created on a node
+
+      Scenario: test sc with volgroup present on the node
+          Given a sc is created with volgroup as "lvmvg"
+          And pvc is created referencing the same sc
+          Then lvmvolume cr should get created
+          And cr state should be Ready
```
Abhinandan-Purkait: Is it worth using some application or it's maybe fine as this focuses on the vg configs.

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,131 @@
...
+  
+  Background:
+      Given a 3 lvmnode cluster is configured
+      And each node <node> has <vg> created with same size 
+
+      Scenario: test sc with volgroup present only on one node
+          Given a sc is created with volgroup as "lvmvg1" with no topology
+          When pvc is created referencing the same sc
+          Then pvc should be in Pending state
+          And lvmvolume cr should not be created
```
Abhinandan-Purkait, line 63: Well it might not always be pending? What if it selects the node that has the vg?

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,131 @@
...
+          When a pvc is created by referencing the sc
+          Then pvc should not be Bound
+          And lvmvolume should not be created
+
+  
+      Scenario: test pvc with higher size then vg free space
+          Given "lvmvg" has lesser free space then "lvmvg1"
+          And sc is created with volgroup as "lvmvg"
+          When a pvc is created with size higher then "lvmvg" and lower then "lvmvg1" free space
+          Then pvc should not be boun
```
Abhinandan-Purkait: boun --> bound

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,131 @@
...
+        | node2        | lvmvg       |  1.5G    |
+        | node3        | lvmvgdiff   |  1G      |
+
+        
+      Scenario: test sc with volgroup present on 2 nodes with less space and allowed topology
+          Given an sc is created with volgroup "lvmvg"
+          And node1 and node2 is in allowed topology
+          When a pvc is created with "1.3G" size
+          Then pvc should be Pending
+          And lvmvolume should not be created
```
Abhinandan-Purkait: How about using a pattern here and make the space on some vg more than needed to see if we select the vg which has enough space?

tests/bdd/volgroup_vgpattern_bdd.feature
Abhinandan-Purkait: We need a variation for this with vgpattern?

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: This just focuses on scheduling. However,  I have added one case for testing `WairForFirstConsumer`. That will require a pod.

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: I thought so initially... However, when i was testing manually. It always selects a node which doesnt have matching vg. In between retires of i created vg on node when it failed in the previous attemt. It doesnt pick that node.

I added this case with expectation that it will fail sometimes. But doesnt seem like happening. If it does also i think let it Fail. That would be an inconsistent behaviour.

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: We could, However, we have cases here where there is lots of vgs with carying size. Creating so many PVs with allowed topology just to build the setup seemed cumbersome. Creating a device of specific size and VG on that is easier.

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
+
+Feature: Validate vgpattern and volgroup based scheduling.
+  Background:
+      Given Single lvmnode cluster is configured
+      And lvm vg "lvmvg" is created on a node
+
+      Scenario: test sc with volgroup present on the node
+          Given a sc is created with volgroup as "lvmvg"
+          And pvc is created referencing the same sc
```
dsharma-dc, line 8: let's use full forms of these abbreviations throughout `sc -> storage class` , `cr -> custom resource`, `pvc -> volume`

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
+
+Feature: Validate vgpattern and volgroup based scheduling.
+  Background:
+      Given Single lvmnode cluster is configured
+      And lvm vg "lvmvg" is created on a node
+
+      Scenario: test sc with volgroup present on the node
+          Given a sc is created with volgroup as "lvmvg"
```
dsharma-dc, line 7:
```suggestion
          Given a sc is created with volgroup parameter as "lvmvg"
```

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
...
+          And pvc is created referencing the same sc
+          Then lvmvolume cr should get created
+          And cr state should be Ready
+
+      Scenario: test sc with volgroup not present on the node
+          Given a sc is created with volgroup as "lvmvgn"
```
dsharma-dc, line 13:
```suggestion
          Given a sc is created with volgroup parameter as "lvmvgn"
```

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
...
+          Then pvc should be in Pending
+          And lvmvolume cr should not get created
+
+  
+  Background:
+      Given Single lvmnode cluster is configured
+      And lvm vg "lvmvg" and "lvmvg1" is created on a node
+
+      Scenario: test sc with vgpattern matching atleast one vg
+          Given sc is created with vgpattern "lvmvg1.*"
```
dsharma-dc: The dot `.` will not match the vg on the node I think. It has to be `lvmvg1*`

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
...
+          When a pvc is created by referencing the sc
+          Then pvc should be Bound
+          And lvmvolume should be in "lvmvg1" vg
+
+  
+      Scenario: test sc with vgpattern matching no vg
+          Given sc is created with vgpattern "lvmvgn.*"
+          When a pvc is created by referencing the sc
+          Then pvc should not be Bound
+          And lvmvolume should not be created
```
dsharma-dc:
```suggestion
          And lvmvolume custom resource should not be created
```

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
...
+
+  
+      Scenario: test pvc with higher size then vg free space
```
dsharma-dc: `higher` -> `larger`

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
...
+          Given "lvmvg" has lesser free space then "lvmvg1"
+          And sc is created with volgroup as "lvmvg"
+          When a pvc is created with size higher then "lvmvg" and lower then "lvmvg1" free space
+          Then pvc should not be bound
+          And lvmvolume cr should not be created
```
dsharma-dc: The mention of lvmvg1 is irrelevant here since sc will mention `lvmvg` and pvc just needs to be larger than free space on `lvmvg`

tests/bdd/volgroup_vgpattern_bdd.feature
dsharma-dc: Are we discouraging users via documentation to not use vgpattern and volgroup both in the the storage class. We should do that I think even though we somehow handle both being present in sc.

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,156 @@
...
+          When pvc is created referencing the same sc
+          Then pvc should be bound 
+          And lvmvolume cr should be Ready
+          And lvmvolume node should be allowed topology node
+
+       Scenario: test sc with volgroup not present on any node
+          Given a sc is created with volgroup as "lvmvgn"
+          When pvc is created referencing the same sc
+          Then pvc should be in Pending state
+          And lvmvolume should not be created
```
dsharma-dc:
```suggestion
          And lvmvolume custom resource should not be created
```

tests/bdd/volgroup_vgpattern_bdd.feature
dsharma-dc: We can extend one of these scenario to say: _**when the vg is expanded to 1.5G(or simply say more than the required pvc size), the pvc should become bound**_

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: Currently we have not documented. But we should. There is an enhancement planned to return error during /`CreateVolume` if both are supplied and they dont compliment each other.

Incase both are supplied then volgroup is considered for scheduling.

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: Will do

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: I followed documentation. It does match. tried it manually.

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: I wanted to emphasize that there is another vg on the node which would not match pattern but has more free space to accomodate the pvc requested size..

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: Okay....will  keep pvc as is, making it volume might not be clearly referencing it to pvc.

abhilashshetty04: On the second thought, These all are k8s recognised shortforms. Could be okay to keep all as is??

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: Done

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: done but using cr

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: I have changed it a little. Please check.

tests/bdd/volgroup_vgpattern_bdd.feature
dsharma-dc: yeah, but bdds are kind of generic and not k8s specific. You can refer maybe Mayastor bdds, if they use similar terminology, let's keep that.

tests/bdd/volgroup_vgpattern_bdd.feature
```diff
@@ -0,0 +1,179 @@
...
+  
+      Examples:
+        |  node        | vg           | vg2           |
+        | node1        | lvmvg11       | lvmvg21      |
+        | node2        | lvmvg12       | lvmvg22      |
+        | node3        | lvmvg13       | lvmvg23      |
+
+  Background:
+    Given a 3 node lvm localpv cluster is configured
+    And each node has vg named "lvmvg"
```
dsharma-dc: we'll keep this vg only on the node having custom label, and not on each node.

dsharma-dc (approved):

tests/bdd/volgroup_vgpattern_bdd.feature
abhilashshetty04: Should that matter? Even if its on other nodes also, We will validate eventually lvmvolume schedules on expected node.

w3aman (approved):

PR #315, opened 2024-06-19
test: create a shared LV for use by two app pods

dsharma-dc:
## Pull Request template

Adds a test for shared volume.

**What this PR does?**:
Creates a LV by specifying `shared` parameter in the storage class so that the volume can be mounted to more than one path.

No

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/tests/provision_test.go
+++ b/tests/provision_test.go
@@ -30,9 +30,11 @@ var _ = Describe("[lvmpv] TEST VOLUME PROVISIONING", func() {
 	})
 })
 
-func deleteAppAndPvc(appname string, pvcname string) {
-	By("Deleting the application deployment")
-	deleteAppDeployment(appName)
+func deleteAppAndPvc(appnames []string, pvcname string) {
+	for _, appName := range appnames {
+		By("Deleting the application deployment " + appName)
+		deleteAppDeployment(appName)
+	}
 	By("Deleting the PVC")
 	deleteAndVerifyPVC(pvcName)
 }
@@ -56,7 +58,7 @@ func fsVolCreationTest() {
 			resizeAndVerifyPVC(false, "10Gi")
 		}
 
-		deleteAppAndPvc(appName, pvcName)
+		deleteAppAndPvc(appNames, pvcName)
 
 		// PV should be present after PVC deletion since snapshot is present
 		By("Verifying that PV exists after PVC deletion")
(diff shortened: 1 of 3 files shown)


tests/utils.go
```diff
@@ -321,137 +343,141 @@ func resizeAndVerifyPVC(shouldPass bool, size string) {
 }
 func createDeployVerifyApp() {
 	ginkgo.By("creating and deploying app pod")
-	createAndDeployAppPod(appName)
+	createAndDeployAppPod(appNames)
 	time.Sleep(30 * time.Second)
 	ginkgo.By("verifying app pod is running", verifyAppPodRunning)
```
niladrih: This log line seems to suggest that it's checking for one app Pod's status. Maybe this would be better?
```suggestion
	ginkgo.By("verifying app pods are running", verifyAppPodRunning)
```

tests/provision_test.go
```diff
@@ -178,6 +193,7 @@ func volumeCreationTest() {
 	By("Running block volume creation test", blockVolCreationTest)
 	By("Running thin volume creation test", thinVolCreationTest)
 	By("Running leak protection test", leakProtectionTest)
+	By("Running shared LV creation test", sharedVolumeTest)
```
niladrih: It is not very clear what this is trying to test. Maybe we could be a bit more verbose on the test title here?
```suggestion
	By("Running LV shared by two pods on the same node test", sharedVolumeTest)
```

dsharma-dc: Updated

tests/utils.go
dsharma-dc: Updated

tests/provision_test.go
```diff
@@ -91,6 +93,21 @@ func blockVolCreationTest() {
 	By("Deleting storage class", deleteStorageClass)
 }
 
+func sharedVolumeTest() {
+	By("Creating shared LV storage class", createSharedVolStorageClass)
+	By("creating and verifying PVC bound status", createAndVerifyPVC)
+	//we use two fio app pods for this test.
+	appNames = append(appNames, "fio-ci-1")
```
abhilashshetty04, line 100: Could we add a new constant appNames?

abhilashshetty04 (approved):

niladrih (approved):

tests/provision_test.go
dsharma-dc: As discussed, since all deploy and verify functions use the same const, it made more sense to reuse the same.

PR #317, opened 2024-06-20
test(ci): adding wrapper for vg and pv operations

abhilashshetty04:
This PR adds required wrapper funcations for PV and VG operations. Its required for new tests to be automated

**What this PR does?**:
Same as above

NA

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Commit has unit tests

### DIFF
--- a/tests/provision_test.go
+++ b/tests/provision_test.go
@@ -190,12 +190,26 @@ func leakProtectionTest() {
 	By("Deleting storage class", deleteStorageClass)
 }
 
+// This acts as a test just to call the new wrapper functions,
+// Once we write tests calling them this will not be required.
+// This doesnt test any Openebs component.
+func lvmOps() {
+	device := createPV(6)
+	createVg("newvgcode3", device)
+	device_1 := createPV(5)
+	extendVg("newvgcode3", device_1)
+	removeVg("newvgcode3")
+	removePV(device)
+	removePV(device_1)
+}
+
 func volumeCreationTest() {
 	By("Running filesystem volume creation test", fsVolCreationTest)
 	By("Running block volume creation test", blockVolCreationTest)
 	By("Running thin volume creation test", thinVolCreationTest)
 	By("Running leak protection test", leakProtectionTest)
 	By("Running shared volume for two app pods on same node test", sharedVolumeTest)
+	By("Running Lvm Ops", lvmOps)
 }
 
 func capacityTest() {
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -19,8 +19,6 @@ package tests
 import (
 	"context"
 	"fmt"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/onsi/ginkgo"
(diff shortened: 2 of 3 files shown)


dsharma-dc:
What is the rationale behind these wrappers? Sometime earlier we discussed that we'll leave these operations as it exists today, until we integrate to a common mechanism to manage lvm including Mayastor LVM backend.

abhilashshetty04:
> What is the rationale behind these wrappers? Sometime earlier we discussed that we'll leave these operations as it exists today, until we integrate to a common mechanism to manage lvm including Mayastor LVM backend.

This is going to be used in the ci tests. We have scenarios where there is multiple vg in the node, there is a scenario to do vgextend ans see provisioning work.

tests/utils.go
```diff
@@ -736,6 +736,98 @@ func enableThinpoolMonitoring() {
...
+
+	args_loop := []string{
+		"losetup", "-d",
+		device,
+	}
+	_, _, err_loop := execAtLocal("sudo", nil, args_loop...)
+	gomega.Expect(err_loop).To(gomega.BeNil(), "loop device remove failed")
+	filename := fmt.Sprintf("/tmp/openebs_ci_disk_%s.img", file_id)
+	args_file := []string{
+		"rm",
```
dsharma-dc: you may want to do `rm -f` here

tests/utils.go
```diff
@@ -736,6 +736,98 @@ func enableThinpoolMonitoring() {
 	gomega.Expect(err).To(gomega.BeNil(), "run lvchange command")
 }
 
+// This creates loopdevice using file_id(for unquiness) and size,
+// Uses the new loop device to create PV. returns loopdevice name to the caller.
+func createPV(size int, file_id string) string {
```
dsharma-dc: I suggest making it more generic like - don't pass any `file_id` string. Assign a random string(using a generator) to concatenate on file name e.g. `openebs_ci_disk_fXVftN.img`
While deleting in `removePV` only pass the loop device name and get the back-file for it to delete the file by filtering using `-O` flag .

tests/utils.go
abhilashshetty04: Done

tests/utils.go
abhilashshetty04: Done. Using timestamp for uniqueness

dsharma-dc (approved):

tests/utils.go
```diff
@@ -776,6 +776,109 @@ func enableThinpoolMonitoring() {
...
+		size_str, filename,
+	}
+	_, _, err := execAtLocal("sudo", nil, args_file...)
+	gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "create device failed")
+	args_loop := []string{
+		"losetup", "-f",
+		filename, "--show",
+	}
+
+	stdout_loop, _, err := execAtLocal("sudo", nil, args_loop...)
```
niladrih: We can run the test using sudo, e.g.: https://github.com/openebs/dynamic-localpv-provisioner/blob/v3.5.x/Makefile#L146

If the test is using a shell script, e.g. test.sh, run it using `sudo bash -c "./test.sh"`

We should not call sudo implicitly like so.

tests/utils.go
niladrih, line 1: nit: you could put these lvm operations in a separate file called lvm_ops.go or lvm.go or disk.go or something like that.

sinhaashish:
can you please squash the relevant commits. I see commits `856c95e83e12a4bdd2cb8d6a5e6b2a1e7ce80db3`  and `f1a328cd2758e5a8f52ca0aabde3abc46a91d73b` are modifying the same file.

tests/provision_test.go
```diff
@@ -190,12 +190,26 @@ func leakProtectionTest() {
 	By("Deleting storage class", deleteStorageClass)
 }
 
+// This acts as a test just to call the new wrapper functions,
+// Once we write tests calling them this will not be required.
+// This doesnt test any Openebs component.
+func lvmOps() {
```
niladrih, line 196: Why is this required? Isn't there a script which sets these up already?

abhilashshetty04: No that will be in next PR.

tests/utils.go
niladrih: Maybe use `mktemp` or Go's Tempfile API, e.g. https://github.com/openebs/dynamic-localpv-provisioner/blob/v3.5.x/tests/suite_test.go#L95?

tests/utils.go
dsharma-dc: The main ci script in lvm repo runs as non-root user in github actions. Most of the work done inside that script is doing `sudo` today e.g. `sudo vgcreate`. These new wrappers seem to be emulating same.

tests/utils.go
abhilashshetty04: Done, moved to lvm_util.go

tests/utils.go
niladrih: This feels like tech debt. But oh well.

tiagolobocastro: We could have a wrapper for calls which require sudo. In general IMHO it's better to use sudo only for operations which need it rather than blindly run everything with sudo as that seems a lot more dangerous.

PR #320, opened 2024-06-26
test(ci): adding volgroup based tests

abhilashshetty04:
**What this PR does?**:
This PR adds new ci tests based on volgroup/vgpattern based scheduling

No

### DIFF
--- a/tests/lvm_utils.go
+++ b/tests/lvm_utils.go
@@ -20,17 +20,20 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
 )
 
 // This creates loopdevice using the size passed as arg,
 // Uses the new loop device to create PV. returns loopdevice name to the caller.
 func createPV(size int) string {
-	fmt.Printf("Creating device\n")
+	ginkgo.By("Creating Pv")
 
 	back_file_args := []string{
-		"mktemp", "-t",
+		"mktemp",
+		"-t",
 		"openebs_lvm_localpv_disk_XXXXX",
 		"--dry-run",
 	}
@@ -39,14 +42,16 @@ func createPV(size int) string {
 	file_str := strings.TrimSpace(string(file[:]))
 	size_str := strconv.Itoa(size) + "G"
 	device_args := []string{
-		"truncate", "-s",
+		"truncate",
+		"-s",
 		size_str, file_str,
 	}
 	_, _, err := execAtLocal("sudo", nil, device_args...)
 	gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "create device failed")
 
 	args_loop := []string{
-		"losetup", "-f",
+		"losetup",
+		"-f",
 		file_str, "--show",
 	}
 	stdout_loop, _, err := execAtLocal("sudo", nil, args_loop...)
@@ -62,9 +67,29 @@ func createPV(size int) string {
 	return stdout_loop_str
 }
 
+// Gets lv_count of a specified vg, returns false if its not empty at the end of poll.
+func vgEmpty(name string) bool {
+	args_lvs := []string{
+		"vgs",
+		name,
+		"--options",
+		"lv_count",
+		"--noheadings",
+	}
+	lvs, _, _ := execAtLocal("sudo", nil, args_lvs...)
+	lvs_str := strings.TrimSpace(string(lvs))
+	lv_cnt, _ := strconv.Atoi(lvs_str)
+	fmt.Printf("lvs cnt is %d\n", lv_cnt)
+	if lv_cnt != 0 {
+		return false
+	} else {
+		return true
+	}
+}
+
 // Does pvremove on specified device. Deletes loop device and the file backing loop device.
 func removePV(device string) {
-	fmt.Printf("remove pv\n")
+	ginkgo.By("Removing pv")
 	args_pv := []string{
 		"pvremove",
 		device,
@@ -84,7 +109,8 @@ func removePV(device string) {
 	dev_str := strings.TrimSpace(string(dev))
 
 	args_loop := []string{
-		"losetup", "-d",
+		"losetup",
+		"-d",
 		device,
 	}
 	_, _, err_loop := execAtLocal("sudo", nil, args_loop...)
@@ -101,7 +127,7 @@ func removePV(device string) {
 
 // Creates vg on the specified device, Device passed should be a pv.
 func createVg(name string, device string) {
-	fmt.Printf("Creating vg\n")
+	ginkgo.By("Creating vg")
 	args_vg := []string{
 		"vgcreate", name,
 		device,
@@ -112,7 +138,7 @@ func createVg(name string, device string) {
 
 // Takes vg name and pv device, extends vg using the supplied pv.
 func extendVg(name string, device string) {
-	fmt.Printf("extending vg\n")
+	ginkgo.By("Extending vg")
 	args_vg := []string{
 		"vgextend", name,
 		device,
@@ -121,14 +147,31 @@ func extendVg(name string, device string) {
 	gomega.Expect(err_vg).To(gomega.BeNil(), "vg extend failed")
 }
 
-// Does vhremove on specified vg with force flag,
-// lv will be forcedeleted if vg is not empty.
+// Does vgremove on specified vg with -y flag if vg isnt empty after fer retires.
 func removeVg(name string) {
-	fmt.Printf("Removing vg\n")
+	ginkgo.By("Removing vg")
+	retries := 3
+	current_retry := 0
 	args_vg := []string{
 		"vgremove",
 		name,
-		"-f",
+	}
+	for {
+		if current_retry < retries {
+			vg_empty := vgEmpty(name)
+			if vg_empty {
+				fmt.Printf("No lv in vg before vg remove\n")
+				break
+			} else {
+				fmt.Printf("lv in vg during retry %d\n", current_retry)
+			}
+		} else {
+			fmt.Printf("vg still not empty after 6 seconds, moving on with force delete\n")
+			args_vg = append(args_vg, "-f")
+			break
+		}
+		current_retry += 1
+		time.Sleep(2 * time.Second)
 	}
 	_, _, err_vg := execAtLocal("sudo", nil, args_vg...)
 	gomega.Expect(err_vg).To(gomega.BeNil(), "vg remove failed")
@@ -137,6 +180,7 @@ func removeVg(name string) {
 // enable the monitoring on thinpool created for test, on local node which
 // is part of single node cluster.
 func enableThinpoolMonitoring() {
+	ginkgo.By("Enable thinpool monitoring")
 	lv := VOLGROUP + "/" + pvcObj.Spec.VolumeName
 
 	args := []string{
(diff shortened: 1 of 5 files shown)


tests/lvm_utils.go
```diff
@@ -20,14 +20,17 @@ import (
...
 
+	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
 )
 
 // This creates loopdevice using the size passed as arg,
 // Uses the new loop device to create PV. returns loopdevice name to the caller.
 func createPV(size int) string {
-	fmt.Printf("Creating device\n")
+	time.Sleep(2 * time.Second)
```
tiagolobocastro: what is the sleep for?

tests/provision_test.go
```diff
@@ -40,13 +40,17 @@ func deleteAppAndPvc(appnames []string, pvcname string) {
...
 func fsVolCreationTest() {
+	device := createPV(20)
+	createVg("lvmvg", device)
 	fstypes := []string{"ext4", "xfs", "btrfs"}
 	for _, fstype := range fstypes {
 		By("####### Creating the storage class : " + fstype + " #######")
 		createFstypeStorageClass(fstype)
-		By("creating and verifying PVC bound status", createAndVerifyPVC)
+		By("creating and verifying PVC bound status")
+		createAndVerifyPVC(true)
```
tiagolobocastro, line 60: what is the bool for?

tests/utils.go
```diff
@@ -242,11 +289,11 @@ func createAndVerifyPVC() {
...
 	)
 
-	ginkgo.By("verifying pvc status as bound")
+	ginkgo.By("verifying pvc status as bound\n")
+	ok := IsPVCBoundEventually(pvcName, 15, 3)
 
-	status := IsPVCBoundEventually(pvcName)
-	gomega.Expect(status).To(gomega.Equal(true),
-		"while checking status equal to bound")
+	gomega.Expect(ok).To(gomega.Equal(expect_bound),
```
tiagolobocastro: Seems the not bound here is simple relying on it not being bound within a timeout?
Could we check for expected errors instead?

tests/provision_test.go
```diff
@@ -70,13 +74,19 @@ func fsVolCreationTest() {
 		verifyPVForPVC(false, pvcName)
 		By("Deleting storage class", deleteStorageClass)
 	}
+	removeVg("lvmvg")
```
tiagolobocastro: this should be done with a defer

tiagolobocastro: or perhaps better to use beforeEach and afterEach kind of thing?

dsharma-dc: I have suggested not doing PV and VG setup for each test. Rather setup once at ginkgo entry point, use for all tests, and cleanup at the end. Only if a test needs anything additional can perform those additional setup from that test.

tests/provision_test.go
abhilashshetty04: We have added some cases where PVC might not get Bound. Previously written cases all were expected to Bound. depending upon this arg. Validation will change.

tests/utils.go
abhilashshetty04: k8s doesnt append any errors in PVC. We can check only Phase (Bound or Pending) to determine the provisioning. We wait for 15 seconds with 3 secs poll. Ive noticed it does 3 to 4 retries within first 15 secs.

tests/lvm_utils.go
abhilashshetty04: I will remove it.

tests/provision_test.go
abhilashshetty04: Yes doing cleanup and setup in the suite level

tests/utils.go
tiagolobocastro: I might be misremembering but isn't there an error message on the pvc events or something like that?

tests/lvm_utils.go
```diff
@@ -124,11 +149,34 @@ func extendVg(name string, device string) {
 // Does vhremove on specified vg with force flag,
```
dsharma-dc:
```suggestion
// Does vgremove on specified vg with force flag,
```

tests/lvm_utils.go
```diff
@@ -124,11 +149,34 @@ func extendVg(name string, device string) {
...
+				fmt.Printf("lv in vg during retry %d\n", current_retry)
+			}
+		} else {
+			fmt.Printf("vg still not empty after 5 seconds, moving on with force delete\n")
+			force = true
+			break
+		}
+		current_retry += 1
+		time.Sleep(2 * time.Second)
+	}
```
dsharma-dc:
```suggestion
  for current_retry < retries {
      vg_empty := vgEmpty(name)
      if vg_empty {
        fmt.Printf("No lv in vg, removing vg %v\n", name)
        break
      } else {
        fmt.Printf("vg has lvs, retrying %d\n", current_retry)
      }

    current_retry += 1
    time.Sleep(2 * time.Second)
  }
```

tests/lvm_utils.go
```diff
@@ -124,11 +149,34 @@ func extendVg(name string, device string) {
...
+
 	args_vg := []string{
 		"vgremove",
 		name,
-		"-f",
+	}
+	if force {
+		args_vg = append(args_vg, "-y")
 	}
```
dsharma-dc:
```suggestion
	if force {
	        fmt.Printf("vg still not empty after 5 seconds, moving on with force delete\n")
		args_vg = append(args_vg, "-y")
	}
```

tests/lvm_utils.go
```diff
@@ -124,11 +149,34 @@ func extendVg(name string, device string) {
...
+		if current_retry < retries {
+			vg_empty := vgEmpty(name)
+			if vg_empty {
+				fmt.Printf("No lv in vg before vg remove\n")
+				break
+			} else {
+				fmt.Printf("lv in vg during retry %d\n", current_retry)
+			}
+		} else {
+			fmt.Printf("vg still not empty after 5 seconds, moving on with force delete\n")
```
dsharma-dc: You seem to be waiting 10secs.

tests/provision_test.go
```diff
@@ -93,13 +109,68 @@ func blockVolCreationTest() {
 	By("Deleting storage class", deleteStorageClass)
 }
 
+func vgExtendTest() {
```
dsharma-dc: May be a good idea to mention bdd scenario name as a function comment here(and in other bdd related tests) for quick reference?

tests/utils.go
```diff
@@ -193,19 +244,43 @@ func createThinStorageClass() {
 }
 
 // VerifyLVMVolume verify the properties of a lvm-volume
-func VerifyLVMVolume() {
+// expected_vg is supposed to be passed only when vgpatten was used for scheduling.
+// If its volgroup in sc then we can just match volgroup with lvmvolume's vg field.
+func VerifyLVMVolume(no_provision bool, expected_vg string) {
```
dsharma-dc:
```suggestion
func VerifyLVMVolume(expect_provisioned bool, expected_vg string) {
```

Accordingly, following checks would need to be 
```
if !expect_provisioned {
} else {
}
```

tests/utils.go
```diff
@@ -41,16 +41,25 @@ import (
...
-func IsPVCBoundEventually(pvcName string) bool {
-	return gomega.Eventually(func() bool {
-		volume, err := PVCClient.
-			Get(pvcName, metav1.GetOptions{})
-		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-		return pvc.NewForAPIObject(volume).IsBound()
-	},
-		60, 5).
-		Should(gomega.BeTrue())
+func IsPVCBoundEventually(pvcName string, timeoutSeconds int, intervalSeconds int) bool {
```
dsharma-dc: Why are we rewriting polling and waiting logic when `gomega.Eventually` does the same. We can pass the `timeoutSeconds` and `intervalSeconds` to `gomega.Eventually` instead of hardcodes `60` and `5`

tests/utils.go
```diff
@@ -146,6 +155,48 @@ func createStorageClass() {
 	gomega.Expect(err).To(gomega.BeNil(), "while creating a default storageclass {%s}", scName)
 }
 
+func createVgPatternStorageClass() {
```
dsharma-dc, line 160: As a separate task, we'll have to consolidate and make a flexible storage class creation function so that we don't write a new function for every new tweak in storage class field.

tests/utils.go
abhilashshetty04: Even i thought it should work, But when pvc is expected to be Pending, gomega logic should return false after exhausting all retires. it returns true if condition match. panics if it doesnt at the end of poll

tests/utils.go
abhilashshetty04: Makes sense.

tests/lvm_utils.go
abhilashshetty04: We want to have a logic when retries exhaused (to mark force: true) with this approach we just exit loop.

tests/lvm_utils.go
abhilashshetty04: making it 6 secs now. 10 seemed too much.

tests/provision_test.go
abhilashshetty04: Done

tests/lvm_utils.go
```diff
@@ -124,11 +149,34 @@ func extendVg(name string, device string) {
...
+		time.Sleep(2 * time.Second)
+	}
+
 	args_vg := []string{
 		"vgremove",
 		name,
-		"-f",
+	}
+	if force {
+		args_vg = append(args_vg, "-y")
```
dsharma-dc: Do you intend to append `-f` here and not `-y` ?

tests/lvm_utils.go
dsharma-dc: My bad, you can still keep the loop simpler and instead of `force` boolean just check:

```
if vg_is_not_empty {
    do_force_removal_now
}
```

tests/lvm_utils.go
abhilashshetty04: -y is yes for all. If lv is in vg while vgremove it asks whether to remove lv. So used -y.

tests/utils.go
abhilashshetty04: We get these events on pvc:

```
Events:
  Type     Reason                Age                From                                                                                               Message
  ----     ------                ----               ----                                                                                               -------
  Normal   ExternalProvisioning  11s (x4 over 24s)  persistentvolume-controller                                                                        Waiting for a volume to be created either by the external provisioner 'local.csi.openebs.io' or manually by the system administrator. If volume creation is delayed, please verify that the provisioner is running and correctly registered.
  Warning  ProvisioningFailed    4s (x5 over 21s)   local.csi.openebs.io_openebs-lvm-controller-546768b57c-drszh_3c77800c-663f-469c-a3e5-be02414c02b8  failed to provision volume with StorageClass "openebs-lvmpv-non": rpc error: code = ResourceExhausted desc = no vg available to serve volume request having regex="^foo1n$" & capacity="4294967296"
  Normal   Provisioning          1s (x6 over 24s)   local.csi.openebs.io_openebs-lvm-controller-546768b57c-drszh_3c77800c-663f-469c-a3e5-be02414c02b8  External provisioner is provisioning volume for claim "default/csi-lvmpv"
```

However, im thinking just checking PVC state to be not Bound should suffice. We also validate lvmvolume cr for negative scenario after this.

Moreover, plugin appends same error on events incase vg present but cant fulfil capacity request also.

tests/utils.go
abhilashshetty04: Done

tests/lvm_utils.go
abhilashshetty04: ill use -f

tests/utils.go
dsharma-dc: As discussed, we'll go ahead writing a separate `IsPvcPendingConsistently` function using the gomega `Consistently` checker for negative cases.

dsharma-dc (approved):

PR #323, opened 2024-07-09
test(bdd): adding scheduler logic bdd

abhilashshetty04:


tests/bdd/scheduler.feature
```diff
@@ -0,0 +1,53 @@
...
+            Scenario: test CapacityWeighted scheduler logic
+                Given sc is created with scheduler parameter set to "CapacityWeighted"
+                When "pvc-1" is created using size "5G"
+                And "pvc-2" is created using size "5G"
+                And "pvc-3" is created using size "2G"
+                Then all three pvc lvmvolume should be placed on different nodes
+                When "pvc-4" is created with size "1G"
+                Then "pvc-4" lvmvolume should be placed on "pvc-3" lvmvolume node
+                When "pvc-5" is created with size "1G"
+                Then "pvc-5" lvmvolume should be placed on "pvc-4" lvmvolume node
```
Abhinandan-Purkait: pvc-4 --> pvc-3, let's use the same reference node, as that helps in automation

tests/bdd/scheduler.feature
```diff
@@ -0,0 +1,53 @@
...
+                When "pvc-6" is created with size "2G"
+                Then "pvc-6" lvmvolume should be placed on "pvc-5" lvmvolume node
```
Abhinandan-Purkait: pvc-5 --> pvc-3

tests/bdd/scheduler.feature
```diff
@@ -0,0 +1,53 @@
...
+
+            Scenario: test SpaceWeighted with vgextend logic
+                Given sc is created by not setting scheduler parameter explicitly
+                When "pvc-1" is created with using size "6G"
+                And "pvc-2" is created with using size "2G"
+                And "pvc-3" is created with using size "3G"
+                Then all pvc lvmvolume should be placed on different nodes
+                When "pvc-1" vg is extended by 10G
+                And "pvc-4" is created with size "5G" 
+                Then "pvc-4" lvmvolume should be placed on "pvc-1" lvmvolume node
```
Abhinandan-Purkait, line 31: Should it not go to pvc-2 node?

Abhinandan-Purkait (commented):
Can you please explain what is the difference between SpaceWeighted and CapacityWeighted?

abhilashshetty04:
> Can you please explain what is the difference between SpaceWeighted and CapacityWeighted?

In `SpaceWeighted`, scheduler picks node which has vg with larger free space, Whereas in `CapacityWeighted` it picks node with vg that has least allocated capacity.

In `VolumeWeighted` scheduler will pick a node with vg having less number of lv.

tests/bdd/scheduler.feature
abhilashshetty04: No We did `vgextend` on pvc-1 node right? So after extend `pvc-1 `node will have more free space. Next case is similar without vgextend , In that we expect it to go to `pvc-2 `node

tests/bdd/scheduler.feature
```diff
@@ -0,0 +1,53 @@
...
+                When "pvc-4" is created with size "1G"
+                Then "pvc-4" lvmvolume should be placed on "pvc-3" lvmvolume node
+                When "pvc-5" is created with size "1G"
+                Then "pvc-5" lvmvolume should be placed on "pvc-3" lvmvolume node
+                When "pvc-6" is created with size "2G"
+                Then "pvc-6" lvmvolume should be placed on "pvc-3" lvmvolume node
+                When "pvc-7" is created with size "3G"
+                Then "pvc-7" lvmvolume should be placed on "pvc-1" or "pvc-2" lvmvolume node
+                When "pvc-8" is created with size "3G"
+                Then "pvc-8" lvmvolume should not be placed on "pvc-7" and "pvc-3" node
```
dsharma-dc, line 21: Do we really need these many volumes to be created? Instead add a case where we free some space on a node and then expect a new volume scheduled there.

tests/bdd/scheduler.feature
```diff
@@ -0,0 +1,53 @@
...
+            Scenario: test SpaceWeighted without vgextend logic
+                Given sc is created by not setting scheduler parameter explicitly
+                When "pvc-1" is created with using size "6G"
+                And "pvc-2" is created with using size "2G"
+                And "pvc-3" is created with using size "3G"
+                Then all pvc lvmvolume should be placed on different nodes
+                When "pvc-4" is created with size "5G" 
+                Then "pvc-4" lvmvolume should be placed on "pvc-2" lvmvolume node
+
+            Scenario: test VolumeWighted logic
```
dsharma-dc:
```suggestion
            Scenario: test VolumeWeighted logic
```

dsharma-dc (commented):
Instead of referring nodes like `pvc-1 node` or `pvc-2 node` etc, better define node names `node1` `node2` and `node3` and use them as reference.

tests/bdd/scheduler.feature
abhilashshetty04: This suite is to test volume scheduling over multiple provisioning. Every volume creation defines where the next provisioning will happen. Even the size of the volume matters as its `CapacityWeighted` test. 

W.r.t Automation also this shouldnt take much time. Its 8 volume in total.

abhilashshetty04:
> Instead of referring nodes like `pvc-1 node` or `pvc-2 node` etc, better define node names `node1` `node2` and `node3` and use them as reference.

We dont know where the initial provisioning will happen. But we can predict which volume a new volume could follow. So i thought of validating with pvc-x volume node.

It could be confusing if we specify node name.

Abhinandan-Purkait (approved):

dsharma-dc (approved):

PR #324, opened 2024-07-23
ci: enhance pull_request workflow

dsharma-dc:
### Workflow in detail

1. A user creates a PR to develop or release/x.y branch.
2. The concerned GitHub action workflow kicks in on the PR.
3. First, it checks out and sets up the runner.
4. It then runs the lint. Shellcheck and helm lint.
5. Verify the chart version, it should be of the format x.y+1.0-develop. If it’s anything else, the workflow fails. From now on any change to the helm chart in the PR would not need a helm version change. Users are not supposed to change the chart version in their PRs, even if it’s a helm change.
6. Extract the chart version from the chart. This tag would be used for building the image locally.
7. Build the images using the code in the branch and tag it with x.y+1.0-develop
8. Set up the minikube cluster.
9. Execute the CI tests by installing the helm chart from the deploy/helm/charts directory. Set the image pull policy to Never since we are supposed to use the images built in the previous step.
10. Execute the CI tests. We are to use the helm chart here.
11. If the test passes the CI succeeds else it fails.


.github/workflows/pull_request.yml
```diff
@@ -62,7 +49,19 @@ jobs:
           path: '.'
           pattern: '*.sh'
 
-  unit-test:
+      - uses: cachix/install-nix-action@v22
+      - uses: rrbutani/use-nix-shell-action@v1.1.0
+        with:
+            file: shell.nix
+      - name: Check if the chart follows a valid semver version
```
Abhinandan-Purkait, line 57: Add a line above

.github/workflows/pull_request.yml
```diff
@@ -62,7 +49,19 @@ jobs:
...
+        run: |
+          branch_name="${{ github.event.pull_request.base.ref }}"
+          ./scripts/validate-chart-version.sh --branch $branch_name
+  
+      - name: Run chart-testing lint
+        run: |
+          ct lint --config ct.yaml
+  unit-tests:
```
Abhinandan-Purkait, line 66: Add a line above

Abhinandan-Purkait (approved):

.github/workflows/pull_request.yml
dsharma-dc: done

dsharma-dc:
Any more reviews on this please?

pchandra19 (approved):

abhilashshetty04 (approved):

PR #333, opened 2024-09-19
fix(chart): revert env OPENEBS_NAMESPACE to LVM_NAMESPACE for v1.6.x

niladrih:
The env LVM_NAMESPACE was changed to OPENEBS_NAMESPACE on the develop branch as a part of https://github.com/openebs/lvm-localpv/pull/324. This change breaks v1.6.x.

Need to investigate if it breaks develop as well.

### DIFF
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: lvm-localpv
 description: CSI Driver for dynamic provisioning of LVM Persistent Local Volumes.
-version: 1.6.1
+version: 1.6.2
 appVersion: 1.6.1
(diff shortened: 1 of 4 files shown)


deploy/helm/charts/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: lvm-localpv
 description: CSI Driver for dynamic provisioning of LVM Persistent Local Volumes.
-version: 1.6.1
+version: 1.6.2
 appVersion: 1.6.1
```
avishnu, line 5: Shouldn't the app version also change?

niladrih: No, the code sees no change, the appVersion (and container image version) are the same. Changes are going into the helm chart exclusively.

avishnu (approved):

w3aman (approved):

PR #352, opened 2024-12-10
feat(chores): Add missing tools to nix-shell

mhkarimi1383:
## Pull Request template

**Why is this PR required? What issue does it fix?**: controller-tools and some other tools are missing from `nix-shell`

**What this PR does?**: adds corresponding packages

**Does this PR require any upgrade changes?**: No

just run `nix-shell`

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests
- [x] (Optional) Are upgrade changes included in this PR? If not, mention the issue/PR to track:
- [x] (Optional) If documentation changes are required, which issue on https://github.com/openebs/openebs-docs is used to track them:


mhkarimi1383:
I also want to add .envrc to automatically load that

tiagolobocastro:
Pardon my ignorance, what's in `kubernetes-controller-tools` ?

mhkarimi1383:
@tiagolobocastro 
controller-gen cli

tiagolobocastro (approved):

dsharma-dc:
Though nothing harmful, do we really need this in product nix shell ? The required controller-gen is currently installed from the Makefile in the repo and is used to generate manifests.

mhkarimi1383:
In my NixOS I ran `make -j 1` and controller-gen not found and with the latest one from nix repos (unstable) I got unrecognized option error for controller-gen command

tiagolobocastro:
> Though nothing harmful, do we really need this in product nix shell ? The required controller-gen is currently installed from the Makefile in the repo and is used to generate manifests.

Downloaded binaries are not likely to work on nix and NixOS, may require patching et all.
It's simpler to just put it on the nix-shell rather than figure out why it doesn't work.

dsharma-dc (approved):

mhkarimi1383:
I made this PR draft since I see that some more tools are missing....

I want to add every required tools to shell.nix to make easier to work for NixOS users :)

mhkarimi1383:
Commit notes/descriptions are important

mhkarimi1383:
Also we can change all of the actions to use nix shell, so changing go version, etc., will be easier

tiagolobocastro:
Sounds good, just let us know when this is working and ready to merge 👍

mhkarimi1383:
@tiagolobocastro 
Looks like ready, But since I'm getting https://github.com/openebs/lvm-localpv/issues/353 error I can't be sure about that

coverage.txt
dsharma-dc, line 1: If I'm not mistaken, this is an autogenerated file and doesn't need to be checked-in.

mhkarimi1383: Yes, It should be in `.gitignore` :)
I have pushed that file by mistake

mhkarimi1383:
Almost complete,
I have isolated nix-shell's gopath and tempdir, Since installing specific version of some packages is a bit messy

shell.nix
```diff
@@ -5,8 +5,24 @@ in
...
   ];
-}
\ No newline at end of file
+  shellHook = ''
+    export GOPATH=$(pwd)/nix/.go
+    export GOCACHE=$(pwd)/nix/.go/cache
+    export TMPDIR=$(pwd)/nix/.tmp
+    export PATH=$GOPATH/bin:$PATH
+    mkdir -p "$TMPDIR"
+    go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0
```
tiagolobocastro: I think this is now the same the Makefile is installing? Do we need to do it here?

mhkarimi1383: But setting env variables was required, so I can run `make bootstrap` instead of `go install`

tiagolobocastro: Yeah that's fine, just wondering about specifically the install to avoid having to modify this file once we update the Makefile

tiagolobocastro: That said we also have the specific go version here, so we might need to change anyway :)

mhkarimi1383: Yeah, Also if we upgrade nixpkgs specified go version will not be available

mhkarimi1383: I would recommend changing the entire CI system to use nix-shell just like lint action

With that we will have a single source of truth :)

mhkarimi1383: Switched to `make bootstrap`, but getting

```
go: github.com/axw/gocov/gocov@latest (in github.com/axw/gocov@v1.2.1): go.mod:3: invalid go version '1.22.0': must match format 1.23
go.mod:5: unknown directive: toolchain
```

in the logs

but tools are installed correctly as far as I can tell

mhkarimi1383: Also getting

```
  go: gopkg.in/matm/v1/gocov-html@latest: unrecognized import path "gopkg.in": parse https://gopkg.in/?go-get=1: no go-import meta tags ()
```


Fixed in https://github.com/openebs/lvm-localpv/pull/354

shell.nix
```diff
@@ -5,8 +5,25 @@ in
...
+    gnumake
+    minikube
     semver-tool
     yq-go
-    chart-testing
   ];
-}
\ No newline at end of file
+  shellHook = ''
+    export HOME=${builtins.getEnv "HOME"}
```
tiagolobocastro, line 21: why is this needed?

mhkarimi1383: In GitHub actions it's required to set HOME specifically, it will fail if you don't set that (in both action file and shell.nix)

mhkarimi1383: You can checkout previous commits action runs

dsharma-dc (approved):

shell.nix
tiagolobocastro: Interesting, was not aware of this :+1:

tiagolobocastro (approved):

PR #335, opened 2024-09-19
feat(provisioning): extra format options (`mkfs`) added

mhkarimi1383:
## Pull Request template

Please, go through these steps before you submit a PR.

**Why is this PR required? What issue does it fix?**:

adding extra format (`mkfs`) options to newly created Volumes

also switched from deprecated `utils/monut` to `mount-utils` to be able to add this feature and moving from an old package to a maintained one 

**What this PR does?**:

Refactor to new mount utils package and change format and mount function

**Does this PR require any upgrade changes?**:

No, But requires document updates for StorageClasses

**If the changes in this PR are manually verified, list down the scenarios covered:**:

**Any additional information for your reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [x] Fixes #296
- [x] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [x] Has the change log section been updated?
- [ ] Commit has unit tests
- [x] Commit has integration tests
- [ ] (Optional) Are upgrade changes included in this PR? If not, mention the issue/PR to track:
- [x] (Optional) If documentation changes are required, which issue on https://github.com/openebs/openebs-docs is used to track them:

PR #360, opened 2025-01-05
build: a number of fixes on Makefile and nix-shell

tiagolobocastro:
**What this PR does?**:
Ensure the namespace is always set correctly to openebs by default.
Set the image pull policy properly (wrong var was used...)

The non-buildx Dockerfile copies a binary built outside of the docker
build, which may be broken if the dependencies are not met.
This happens on nix-shell as the deps exist on /nix/store.
Get around this by building static binaries in the nix-shell.

Pin a few packages version which are not compatible on latest.
Perhaps all of them should be pinned!
Remove a few uses of sudo which should not be required?
Add a few packages missing from nix-shell
Allow make clean to work properly
Don't bootstrap by default, though this may be discussed further.
Setup goenv env variables on github actions workflows
Add gitignore for the coverage data

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,7 @@ clean:
 	@echo "--> Cleaning Directory" ;
 	go clean -testcache
 	rm -rf bin
+	CLEANUP_ONLY=1 ./ci/ci-test.sh
 	chmod -R u+w ${GOPATH}/bin/${CSI_DRIVER} 2>/dev/null || true
 	chmod -R u+w ${GOPATH}/pkg/* 2>/dev/null || true
 	rm -rf ${GOPATH}/bin/${CSI_DRIVER}
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,7 @@ install-golangci-lint:
 
 .PHONY: controller-gen
 controller-gen:
-	TMP_DIR=$(shell mktemp -d) && cd $$TMP_DIR && go mod init tmp && go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0 && rm -rf $$TMP_DIR;
+	@go install -mod=mod sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0
 
 # SRC_PKG is the path of code files
 SRC_PKG := github.com/openebs/lvm-localpv/pkg
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,10 @@ VETARGS?=-asmdecl -atomic -bool -buildtags -copylocks -methods \
 EXTERNAL_TOOLS=\
 	golang.org/x/tools/cmd/cover \
 	golang.org/x/lint/golint \
-	github.com/axw/gocov/gocov \
+	github.com/axw/gocov/gocov@v1.1 \
 	github.com/matm/gocov-html/cmd/gocov-html \
 	github.com/onsi/ginkgo/ginkgo \
-	github.com/onsi/gomega/...
+	github.com/onsi/gomega/...@v1.35
 
 # The images can be pushed to any docker/image registeries
 # like docker hub, quay. The registries are specified in
@@ -86,6 +86,8 @@ clean:
 	@echo "--> Cleaning Directory" ;
 	go clean -testcache
 	rm -rf bin
+	chmod -R u+w ${GOPATH}/bin/${CSI_DRIVER} 2>/dev/null || true
+	chmod -R u+w ${GOPATH}/pkg/* 2>/dev/null || true
 	rm -rf ${GOPATH}/bin/${CSI_DRIVER}
 	rm -rf ${GOPATH}/pkg/*
 
@@ -119,7 +121,10 @@ verify-deps: deps
 bootstrap: controller-gen install-golangci-lint
 	@for tool in  $(EXTERNAL_TOOLS) ; do \
 		echo "+ Installing $$tool" ; \
-		cd && GO111MODULE=on go install $$tool@latest; \
+		if ! echo $$tool | grep "@"; then \
+			tool=$$tool@latest ; \
+		fi ; \
+		GO111MODULE=on go install $$tool; \
 	done
 
 ## golangci-lint tool used to check linting tools in codebase
@@ -130,7 +135,7 @@ bootstrap: controller-gen install-golangci-lint
 ## Install golangci-lint only if tool doesn't exist in system
 .PHONY: install-golangci-lint
 install-golangci-lint:
-	$(if $(shell which golangci-lint), echo "golangci-lint already exist in system", (curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sudo sh -s -- -b "${GOPATH}/bin" v1.56.2))
+	$(if $(shell which golangci-lint), echo "golangci-lint already exist in system", (curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "${GOPATH}/bin" v1.56.2))
 
 .PHONY: controller-gen
 controller-gen:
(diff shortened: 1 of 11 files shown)


shell.nix
```diff
@@ -16,14 +16,16 @@ pkgs.mkShell {
...
+    util-linux
+  ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure") docker;
   shellHook = ''
-    export HOME=${builtins.getEnv "HOME"}
     export GOPATH=$(pwd)/nix/.go
     export GOCACHE=$(pwd)/nix/.go/cache
     export TMPDIR=$(pwd)/nix/.tmp
     export PATH=$GOPATH/bin:$PATH
     mkdir -p "$TMPDIR"
-    make bootstrap
```
tiagolobocastro, line 27: What do you think @mhkarimi1383? Should we keep this automatically?

tiagolobocastro:
Ah looks like CI failed because GOPATH is not set or empty. Is this expected?
CC @dsharma-dc @Abhinandan-Purkait

tiagolobocastro:
I'm also a bit confused with the `bdd-tests` on PR, it seems like it's testing against develop rather than the code built in this PR or am I'm missing something?

shell.nix
mhkarimi1383: shell.nix file made to make a development environment automatically for Nix users, So I think we should keep this to run automatically it will do nothing and just install requirements like other requirements that we are installing in the same file

mhkarimi1383: shell.nix file made to make a development environment automatically for Nix users, So I think we should keep this to run automatically it will do nothing and just install requirements like other requirements that we are installing in the same file

shell.nix
```diff
@@ -16,14 +16,16 @@ pkgs.mkShell {
...
     semver-tool
     yq-go
-  ];
+    which
+    curl
+    cacert
+    util-linux
+  ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure") docker;
   shellHook = ''
-    export HOME=${builtins.getEnv "HOME"}
```
mhkarimi1383, line 21: HOME env variable should be inherited explicitly when we are using this in GitHub Actions

mhkarimi1383:
> Ah looks like CI failed because GOPATH is not set or empty. Is this expected?
> CC @dsharma-dc @Abhinandan-Purkait 

I think that's because HOME env inheritance is removed

mhkarimi1383:
Can you make builds static?
Because docker builds are not working in Nix and some OSes (Because of shared C libs)

See
https://github.com/openebs/lvm-localpv/pull/335#issuecomment-2564656408

Makefile
```diff
@@ -222,15 +226,15 @@ lvm-driver-image: lvm-driver
...
+	cd buildscripts/${CSI_DRIVER} && docker build -t ${IMAGE_ORG}/${CSI_DRIVER}:${IMAGE_TAG} ${DBUILD_ARGS} . && docker tag ${IMAGE_ORG}/${CSI_DRIVER}:${IMAGE_TAG} quay.io/${IMAGE_ORG}/${CSI_DRIVER}:${IMAGE_TAG}
 	@rm buildscripts/${CSI_DRIVER}/${CSI_DRIVER}
 
 .PHONY: ansible-runner-image
 ansible-runner-image:
 	@echo "------------------"
 	@echo "--> Build ansible-runner image for lvm-localpv e2e-tests" 
 	@echo "------------------"
-	sudo docker build . -f e2e-tests/Dockerfile -t ${IMAGE_ORG}/lvm-localpv-e2e:ci
+	docker build . -f e2e-tests/Dockerfile -t ${IMAGE_ORG}/lvm-localpv-e2e:ci
```
Abhinandan-Purkait: Why do we need the ansible-runner-image? We have migrated the e2e right?

Abhinandan-Purkait:
> Ah looks like CI failed because GOPATH is not set or empty. Is this expected? CC @dsharma-dc @Abhinandan-Purkait

The setup go action step sets it to home/runner/go

Abhinandan-Purkait:
> I'm also a bit confused with the `bdd-tests` on PR, it seems like it's testing against develop rather than the code built in this PR or am I'm missing something?

Which bit are you talking about? It checkout the PR code and build image for it and then uses it from the built image locally?

Makefile
tiagolobocastro: Not sure, that was already there...
I can remove it?

tiagolobocastro:
> > Ah looks like CI failed because GOPATH is not set or empty. Is this expected?
> > CC @dsharma-dc @Abhinandan-Purkait
> 
> I think that's because HOME env inheritance is removed

No: https://github.com/openebs/lvm-localpv/actions/runs/12331398437/job/34417991555#step:7:34

Ah I wanted to ask you, why is this needed? In my case it's always there, even without this line.

tiagolobocastro:
> > Ah looks like CI failed because GOPATH is not set or empty. Is this expected? CC @dsharma-dc @Abhinandan-Purkait
> 
> The setup go action step sets it to home/runner/go

Fixed by taking go env, PTAL

tiagolobocastro:
> > I'm also a bit confused with the `bdd-tests` on PR, it seems like it's testing against develop rather than the code built in this PR or am I'm missing something?
> 
> Which bit are you talking about? It checkout the PR code and build image for it and then uses it from the built image locally?

Discussed offline, I will update here...
So on GHA minikube is setup with driver none, and thus the images are picked up from the host this way.. this is confusing and should be documented.
Also this makes local testing on custom cluster not very straightforward with current script

tiagolobocastro:
> Can you make builds static? Because docker builds are not working in Nix and some OSes (Because of shared C libs)
> 
> See [#335 (comment)](https://github.com/openebs/lvm-localpv/pull/335#issuecomment-2564656408)

The problem here is that the dockerfile is simply copying the binary to the container. On non-nix binaries this might work though it might also fail at runtime potentially if the OS libs are not compatible with the container libs.

On the other hand the buildx is building inside the docker build, which is a bit better, though it seems that the same problem may arise from what I can tell looking at it? 

I'll make use of musl-gcc on the nix-shell and make build static just for the nix-shell for now? Or shall we make it static for docker buildx as well? CC @Abhinandan-Purkait @dsharma-dc

mhkarimi1383:
@tiagolobocastro 

Yes, I think RHEL Based users may face the same problem, NixOS is a big example since it's so different that other distributions in terms of paths, etc.

I think we can also change our flow to build directly through the Dockerfile

(For example if we are running our runtime in debian we can also use golang debian based image)

mhkarimi1383:
With static builds we could also publish binary builds later on (I know the usage for that is rare, but we can do it too)

tiagolobocastro:
A ton of shellcheck warnings were given because I modified the shell file, this is not great to be honest, now it's messed up this PR with a ton of comments.
It would be better to get these in the action, not as comments... -.-

dsharma-dc:
> > Can you make builds static? Because docker builds are not working in Nix and some OSes (Because of shared C libs)
> > See [#335 (comment)](https://github.com/openebs/lvm-localpv/pull/335#issuecomment-2564656408)
> 
> The problem here is that the dockerfile is simply copying the binary to the container. On non-nix binaries this might work though it might also fail at runtime potentially if the OS libs are not compatible with the container libs.
> 
> On the other hand the buildx is building inside the docker build, which is a bit better, though it seems that the same problem may arise from what I can tell looking at it?
> 
> I'll make use of musl-gcc on the nix-shell and make build static just for the nix-shell for now? Or shall we make it static for docker buildx as well? CC @Abhinandan-Purkait @dsharma-dc

If we want to make static for nix as well as docker build, then better do that across all localpv repos if they are also following same?

Abhinandan-Purkait:
> > Can you make builds static? Because docker builds are not working in Nix and some OSes (Because of shared C libs)
> > See [#335 (comment)](https://github.com/openebs/lvm-localpv/pull/335#issuecomment-2564656408)
> 
> The problem here is that the dockerfile is simply copying the binary to the container. On non-nix binaries this might work though it might also fail at runtime potentially if the OS libs are not compatible with the container libs.
> 
> On the other hand the buildx is building inside the docker build, which is a bit better, though it seems that the same problem may arise from what I can tell looking at it?
> 
> I'll make use of musl-gcc on the nix-shell and make build static just for the nix-shell for now? Or shall we make it static for docker buildx as well? CC @Abhinandan-Purkait @dsharma-dc

Yeah, I think that's a better way in general. If we can do it across all repos.

tiagolobocastro:
@Abhinandan-Purkait @dsharma-dc @mhkarimi1383 

Ok, it is done, please give this around review

shell.nix
tiagolobocastro: A valid point, reverted.

dsharma-dc (approved):

Abhinandan-Purkait (approved):

mhkarimi1383:
I have tested the build locally and it was working fine

mhkarimi1383 (approved):

shell.nix
mhkarimi1383: If GitHub Actions are working fine we can ignore this :)

PR #363, opened 2025-01-21
build: various makefile fixes

tiagolobocastro:
build: fix incorrect command args
    
    After adjusting the code, the docs and Makefile were not updated.

---

    build: remove incorrect use of pwd
    
    Makefile commands can be run from another path, example:
    make -C $lvm clean
    
    In this case the pwd is not the correct path prefix!

---
    test: more robust check for lvm driver ready
    
    Ensure there's no deleting pod present
    Also force delete the old pods on uninstall

---

    test(ci): optimize image build and load

### DIFF
--- a/ci/ci-test.sh
+++ b/ci/ci-test.sh
@@ -230,26 +230,52 @@ run() {
   printf "\n\n######### All test cases passed #########\n\n"
 }
 
-load_image() {
-  make lvm-driver-image
+load_k3s() {
   if [ "${CI_K3S:-}" = "true" ]; then
-    docker save openebs/lvm-driver | ctr images import -
+    local img="${1:-}"
+    if [ -z "${1:-}" ]; then
+      repo="$(make image-repo -s -C "$SCRIPT_DIR"/.. 2>/dev/null)"
+      tag="$(make image-tag -s -C "$SCRIPT_DIR"/.. 2>/dev/null)"
+      img="$repo:$tag"
+    fi
+    docker save "$img" | ctr images import -
   fi
 }
 
+load_image() {
+  make lvm-driver-image
+  load_k3s "${1:-}"
+}
+
 maybe_load_image() {
+  local repo tag img did cid
+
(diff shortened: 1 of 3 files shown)


ci/ci-test.sh
```diff
@@ -230,26 +230,52 @@ run() {
...
+      repo="$(make image-repo -s -C "$SCRIPT_DIR"/.. 2>/dev/null)"
+      tag="$(make image-tag -s -C "$SCRIPT_DIR"/.. 2>/dev/null)"
+      img="$repo:$tag"
+    fi
+    docker save "$img" | ctr images import -
   fi
 }
 
+load_image() {
+  make zfs-driver-image
```
dsharma-dc: lvm-driver-image

ci/ci-test.sh
```diff
@@ -230,26 +230,52 @@ run() {
...
-  local id
-  id=$(crictl image --output json | jq --arg image "$(make image-ref -s -C "$SCRIPT_DIR"/.. 2>/dev/null)" '.images[]|select(.repoTags[0] == $image)|.id')
-  if [ -n "$id" ]; then
+  repo="$(make image-repo -s -C "$SCRIPT_DIR"/.. 2>/dev/null)"
+  tag="$(make image-tag -s -C "$SCRIPT_DIR"/.. 2>/dev/null)"
+  img="$repo:$tag"
+
+  did="$(docker image ls --no-trunc --format json | jq -r --arg repo "$repo" --arg tag "$tag" 'select(.Repository == $repo and .Tag == $tag)|.ID')"
+  if [ -z "$did" ]; then
+    make zfs-driver-image
```
dsharma-dc: lvm-driver-image

dsharma-dc (changes requested):

ci/ci-test.sh
tiagolobocastro: doh, fixed

dsharma-dc (approved):

abhilashshetty04 (approved):

Issue #366, opened 2025-02-04
Scheduling Issue: New PVs Scheduled on Cordoned Node


khalMeg:
**Description:**

I am experiencing an unexpected behavior with the scheduling process in openEBS lvm-localpv. Even after cordoning node `worker-06`, new Persistent Volumes (PVs) are still being scheduled on this node.

I was advised by @abhilashshetty04 to create this issue following the discussion in issue [#356](https://github.com/openebs/lvm-localpv/issues/356).

**Steps to Reproduce:**
1. Cordon node `worker-06` using the command:
   ```
   kubectl cordon worker-06
   ```
2. Deploy a workload that requires new PVs.
3. Observe that new PVs are still allocated on `worker-06`.

**Expected Behavior:**
New PVs should not be scheduled on a cordoned node, as it is marked as unschedulable.

**Actual Behavior:**
PVs continue to be provisioned on `worker-06` despite it being cordoned.

**Environment:**
- LVM Driver version
```
LVM version:     2.03.16(2) (2022-05-18)
Library version: 1.02.185 (2022-05-18)
Driver version:  4.47.0
```
- Kubernetes version :
```
Client Version: v1.30.4
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.4
```
- Kubernetes installer & version:
k8s cluster deployed using Kubespray
- OS:
Debian GNU/Linux 12 (bookworm)

I would appreciate any insights or fixes regarding this behavior. Thank you!

abhilashshetty04:
For tracking:
https://github.com/openebs/lvm-localpv/issues/312

PR #375, opened 2025-02-16
exclude cordoned nodes during volume provisioning

abhilashshetty04:
This adds a ci test for cordoned node.

Dependent on:
https://github.com/openebs/lib-csi/pull/26

**Checklist:**
- [X] Fixes #<issue number>
https://github.com/openebs/lvm-localpv/issues/366


tests/utils.go
```diff
@@ -320,6 +322,18 @@ func createAndVerifyPVC(expect_bound bool) {
 	)
 }
 
+// Verifies state of already created pvc based on expect_bound.
+func verifyPVC(pvc_name string, expect_bound bool) {
```
dsharma-dc: Perhaps call this `verifyPVCBoundStatus` , as we are specifically checking only that.

tests/utils.go
```diff
@@ -819,3 +833,80 @@ func createNodeDaemonSet(ds *appsv1.DaemonSet) {
...
+	nodes := listNodes(client)
+	for _, node := range nodes.Items {
+		if !node.Spec.Unschedulable {
+			c_err := cordonNode(client, &node)
+			gomega.Expect(c_err).To(
+				gomega.BeNil(),
+				"Could not cordon node",
+			)
+		}
+
```
dsharma-dc: redundant blank line. Formatting seems a bit off here.

tests/utils.go
dsharma-dc: Also, since this new function is being defined, please remove the same logic from inside `createAndVerifyPVC` and call this new function from there.

dsharma-dc (approved):

Abhinandan-Purkait (approved):

PR #390, opened 2025-05-13
update csi spec version

abhilashshetty04:
Updated csi-spec to v1.11.0.
Updated Go workspace to 1.23.0


abhilashshetty04:
Will continue this : https://github.com/openebs/lvm-localpv/pull/391 without go update

PR #413, opened 2025-09-03
make delete volume idempotent when node is removed from cluster

### LINKED ISSUE #407: Slow leak of LVMVolume Kubernetes objects causes extremely slow allocation of new volumes
**What steps did you take and what happened:**

We have encountered situations where after several weeks of launching pods with LVM volumes, replacing nodes, and generally using the cluster, creation of new volumes slows to a crawl (ie: minutes between creation despite many pods pending and plenty of available space on nodes) and there are hundreds of `LVMVolume` objects with `deletionTimestamp`s and `ownerNodeID`s of non-existent nodes.

Most recently, we found that we had 620 `LVMVolume` objects, with 340 of them in this state. Deleting these leaked `LVMVolume` objects speeds up creation of new ones.

We're not 100% certain, but we believe the leak is caused by nodes getting deleted before the `LVMVolume`s associated with them have been cleaned up.

In our case, we use Karpenter to manage nodes, and it deletes the underlying EC2 instance when the node becomes empty of non-daemonset pods for some period of time.

This may be the same as https://github.com/openebs/lvm-localpv/issues/347, but there aren't enough details in that one to be sure.

**What did you expect to happen:**
1. LVMVolumes on nodes that no longer exist should be deleted, not blocked by a finalizer that never gets removed. If the node doesn't exist, the volume on that node also doesn't exist.
2. Having many LVMVolume objects should not impact creation of new ones.

**The output of the following commands will help us better understand what's going on**:
(Pasting long output into a [GitHub gist](https://gist.github.com) or other [Pastebin](https://pastebin.com/) is fine.)

Controller logs:
```
csi-provisioner I0616 16:21:36.567743       1 event.go:298] Event(v1.ObjectReference{Kind:"PersistentVolume", Namespace:"", Name:"pvc-2c65641e-440e-4e31-8192-1ad6acc240c4", UID:"f09e50ff-c09f-4ded-b65a-1e4c15d23319", APIVersion:"v1", ResourceVersion:"883943724", FieldPath:""}): type: 'Warning' reason: 'VolumeFailedDelete' rpc error: code = DeadlineExceeded desc = context deadline exceeded
csi-provisioner I0616 16:21:36.569874       1 connection.go:200] GRPC response: {}
csi-provisioner I0616 16:21:36.569915       1 connection.go:201] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
csi-provisioner E0616 16:21:36.569952       1 controller.go:1512] delete "pvc-b2317477-1904-4591-98d6-2c5474448c61": volume deletion failed: rpc error: code = DeadlineExceeded desc = context deadline exceeded
csi-provisioner W0616 16:21:36.570012       1 controller.go:989] Retrying syncing volume "pvc-b2317477-1904-4591-98d6-2c5474448c61", failure 1034
csi-provisioner E0616 16:21:36.570030       1 controller.go:1007] error syncing volume "pvc-b2317477-1904-4591-98d6-2c5474448c61": rpc error: code = DeadlineExceeded desc = context deadline exceeded
csi-provisioner I0616 16:21:36.570044       1 event.go:298] Event(v1.ObjectReference{Kind:"PersistentVolume", Namespace:"", Name:"pvc-b2317477-1904-4591-98d6-2c5474448c61", UID:"c8ed79ca-e881-459c-b378-e030e60a4efa", APIVersion:"v1", ResourceVersion:"885072755", FieldPath:""}): type: 'Warning' reason: 'VolumeFailedDelete' rpc error: code = DeadlineExceeded desc = context deadline exceeded
openebs-lvm-plugin I0616 16:21:36.516580       1 controller.go:392] received request to delete volume "pvc-65c90c1a-9eaa-4b29-8a83-7b2a46e4f055"
openebs-lvm-plugin I0616 16:21:36.518581       1 controller.go:392] received request to delete volume "pvc-13b08b9b-6a47-4efb-9062-668e73f62f42"
openebs-lvm-plugin I0616 16:21:36.518835       1 controller.go:392] received request to delete volume "pvc-617ee615-f29b-46dc-8964-b60b50a9c791"
csi-provisioner I0616 16:21:36.735543       1 request.go:628] Waited for 119.260122ms due to client-side throttling, not priority and fairness, request: PATCH:https://10.200.32.1:443/api/v1/namespaces/default/events/pvc-341e0ac9-cada-44f1-9ea8-76bd71876d75.1848b80983569725
openebs-lvm-plugin E0616 16:21:36.534025       1 grpc.go:79] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
csi-provisioner I0616 16:21:36.935593       1 request.go:628] Waited for 192.326659ms due to client-side throttling, not priority and fairness, request: PATCH:https://10.200.32.1:443/api/v1/namespaces/default/events/pvc-2af87e3b-f5d4-4e66-a0bb-4de2e5d2f5b6.1848b8098318babb
csi-provisioner I0616 16:21:37.135645       1 request.go:628] Waited for 191.196251ms due to client-side throttling, not priority and fairness, request: PATCH:https://10.200.32.1:443/api/v1/namespaces/default/events/pvc-f5cb28db-4b43-40b7-a785-93fa31c008de.1848b80e335b53a8
csi-provisioner I0616 16:21:37.335557       1 request.go:628] Waited for 191.260063ms due to client-side throttling, not priority and fairness, request: PATCH:https://10.200.32.1:443/api/v1/namespaces/default/events/pvc-648bb2ba-c626-41c6-abd4-18696886bdbc.1848b80bd7af66f9
openebs-lvm-plugin E0616 16:21:36.547709       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.547809       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.550346       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.550356       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.550364       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.555528       1 grpc.go:79] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
openebs-lvm-plugin E0616 16:21:36.555549       1 grpc.go:79] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
openebs-lvm-plugin E0616 16:21:36.556030       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
csi-provisioner I0616 16:21:37.535169       1 request.go:628] Waited for 191.323743ms due to client-side throttling, not priority and fairness, request: PATCH:https://10.200.32.1:443/api/v1/namespaces/default/events/pvc-2c65641e-440e-4e31-8192-1ad6acc240c4.1848b809831f00f1
csi-provisioner I0616 16:21:37.735129       1 request.go:628] Waited for 191.660542ms due to client-side throttling, not priority and fairness, request: PATCH:https://10.200.32.1:443/api/v1/namespaces/default/events/pvc-b2317477-1904-4591-98d6-2c5474448c61.1848b80e2cce7cc7
openebs-lvm-plugin E0616 16:21:36.558807       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.558762       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.559058       1 grpc.go:79] GRPC error: rpc error: code = Canceled desc = context canceled
openebs-lvm-plugin E0616 16:21:36.564274       1 grpc.go:79] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
csi-provisioner I0616 16:21:40.258526       1 leaderelection.go:276] successfully renewed lease openebs/local-csi-openebs-io
openebs-lvm-plugin E0616 16:21:36.564412       1 grpc.go:79] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
openebs-lvm-plugin E0616 16:21:36.567553       1 grpc.go:79] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
openebs-lvm-plugin E0616 16:21:36.569798       1 grpc.go:79] GRPC error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

We have a script we wrote to clean these leaked volumes up manually.

```python3
#!/usr/bin/env python3
import json
import subprocess
import sys
from argparse import ArgumentParser


def kubectl(
    context: str,
    namespace: str | None,
    args: list[str],
):
    command = ["kubectl", "--context", context]

    if namespace is not None:
        command.extend(["--namespace", namespace])

    command.extend(args)

    return subprocess.run(command, check=True, text=True, capture_output=True)


def parse_args(argv: list[str]):
    parser = ArgumentParser()
    parser.add_argument("--context", type=str, required=True)
    return parser.parse_args(argv)


def main(argv: list[str]):
    args = parse_args(argv)

    lvmvolumes = json.loads(
        kubectl(
            context=args.context,
            namespace="openebs",
            args=["get", "lvmvolumes", "-o", "json"],
        ).stdout,
    )["items"]
    deleting_lvmvolumes = [
        vol for vol in lvmvolumes if "deletionTimestamp" in vol["metadata"]
    ]

    # Intentionally getting node names after lvmvolumes to avoid race conditions
    # where the volume may be created and starting to delete after we query the list of nodes.
    # This relies on only cleaning things with deletionTimestamps to be safe.
    node_names = [
        node["metadata"]["name"]
        for node in json.loads(
            kubectl(
                context=args.context,
                namespace=None,
                args=["get", "nodes", "-o", "json"],
            ).stdout,
        )["items"]
    ]
    for vol in deleting_lvmvolumes:
        if vol["spec"]["ownerNodeID"] not in node_names:
            print(f'Deleting lvmvolume: {vol["metadata"]["name"]}')
            kubectl(
                context=args.context,
                namespace="openebs",
                args=[
                    "patch",
                    "lvmvolumes",
                    vol["metadata"]["name"],
                    "-p",
                    '{"metadata":{"finalizers":null}}',
                    "--type",
                    "merge",
                ],
            )


if __name__ == "__main__":
    main(sys.argv[1:])
```

**Environment:**
- LVM Driver version: v1.6.2
- Kubernetes version (use `kubectl version`): v1.31.6
- Kubernetes installer & version: EKS
- Cloud provider or hardware configuration: AWS r6gd.8xlarge and r6gd.16xlarge
- OS (e.g. from `/etc/os-release`): Bottlerocket 1.39.0-0968c061

abhilashshetty04:
When a Kubernetes node is removed from the cluster while it still has existing PVCs/LVMVolumes, those volumes become orphaned. If a PVC is deleted in this state, the LocalPV-LVM CSI controller marks the corresponding LVMVolume CR for deletion by adding a deletion timestamp.

Normally, the node-agent running on the worker node is responsible for cleaning up the logical volume (LV) and removing the finalizer from the CR. However, in this scenario, the node no longer exists, so there is no agent running to perform the cleanup.

This fix leverages the condition of the Kubernetes Node being absent: in such cases, the CSI controller itself removes the finalizer and cleans up the LVMVolume CR.

Fixes: https://github.com/openebs/lvm-localpv/issues/407

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -19,6 +19,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -471,6 +472,28 @@ func (cs *controller) DeleteVolume(
 	return csipayload.NewDeleteVolumeResponseBuilder().Build(), nil
 }
 
+// Checks if k8s node is present. If the check fails with any other reason then NotFound.
+// We will return true and assume node is present.
+func isNodePresent(nodeId string) bool {
+	cfg, err := k8sapi.Config().Get()
+	if err != nil {
+		return true
+	}
+
+	kubeClient, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return true
+	}
+
+	_, err = kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeId, metav1.GetOptions{})
+	if err != nil {
+		if k8serror.IsNotFound(err) {
+			return false
+		}
+	}
+	return true
+}
+
 func (cs *controller) deleteVolume(ctx context.Context, volumeID string) error {
 	klog.Infof("received request to delete volume %q", volumeID)
 	vol, err := lvm.GetLVMVolume(volumeID)
@@ -481,7 +504,6 @@ func (cs *controller) deleteVolume(ctx context.Context, volumeID string) error {
 		return errors.Wrapf(err,
 			"failed to get volume for {%s}", volumeID)
 	}
-
 	// if volume is not already triggered for deletion, delete the volume.
 	// otherwise, just wait for the existing deletion operation to complete.
 	if vol.GetDeletionTimestamp() == nil {
@@ -489,6 +511,17 @@ func (cs *controller) deleteVolume(ctx context.Context, volumeID string) error {
 			return errors.Wrapf(err,
 				"failed to handle delete volume request for {%s}", volumeID)
 		}
+	}
+	present := isNodePresent(vol.Spec.OwnerNodeID)
+	if !present {
+		klog.Warningf("Removing finalizer as node %s is not present in cluster", vol.Spec.OwnerNodeID)
+		if err = lvm.RemoveVolFinalizer(vol); err != nil {
+			return err
+		}
+		if err = lvm.VerifyLVMVolumeDestroy(volumeID); err != nil {
+			return err
+		}
+		return nil
 	}
 	if err = lvm.WaitForLVMVolumeDestroy(ctx, volumeID); err != nil {
 		return err
--- a/pkg/lvm/volume.go
+++ b/pkg/lvm/volume.go
@@ -104,7 +104,7 @@ func ProvisionVolume(vol *apis.LVMVolume) (*apis.LVMVolume, error) {
 func DeleteVolume(volumeID string) (err error) {
 	err = volbuilder.NewKubeclient().WithNamespace(LvmNamespace).Delete(volumeID)
 	if err == nil {
-		klog.Infof("deprovisioned volume %s", volumeID)
+		klog.Infof("deprovisioning volume %s", volumeID)
 	}
 
 	return
@@ -162,6 +162,23 @@ func WaitForLVMVolumeDestroy(ctx context.Context, volumeID string) error {
 		}
 		timer.Reset(1 * time.Second)
 	}
+}
+
+// Verifies if lvmvolume CR got removed after removing finaliser.
+func VerifyLVMVolumeDestroy(volumeID string) error {
+	for i := 0; i < 10; i++ {
+		_, err := GetLVMVolume(volumeID)
+		if err != nil {
+			if k8serror.IsNotFound(err) {
+				return nil
+			}
+			return status.Errorf(codes.Internal,
+				"lvmvolume: destroy wait failed, not able to get the volume %s %s", volumeID, err.Error())
+		}
+		time.Sleep(time.Duration(100 * time.Millisecond))
+	}
+	return status.Errorf(codes.DeadlineExceeded,
+		"lvmvolume: %s not removed after removing finaliser", volumeID)
 }
 
 // GetLVMVolumeState returns LVMVolume OwnerNode and State for


pkg/lvm/volume.go
```diff
@@ -103,7 +103,7 @@ func ProvisionVolume(vol *apis.LVMVolume) (*apis.LVMVolume, error) {
 func DeleteVolume(volumeID string) (err error) {
 	err = volbuilder.NewKubeclient().WithNamespace(LvmNamespace).Delete(volumeID)
 	if err == nil {
-		klog.Infof("deprovisioned volume %s", volumeID)
+		klog.Infof("deprovisioning volume %s, marking deletion timestamp", volumeID)
```
Abhinandan-Purkait: What do we mean by `marking deletion timestamp`, here?

Abhinandan-Purkait (approved):
LGTM

pkg/driver/controller.go
```diff
@@ -412,6 +434,14 @@ func (cs *controller) deleteVolume(ctx context.Context, volumeID string) error {
 				"failed to handle delete volume request for {%s}", volumeID)
 		}
 	}
+	present := isNodePresent(vol.Spec.OwnerNodeID)
+	if !present {
+		klog.Infof("Removing finalizer as node %s is not present in cluster", vol.Spec.OwnerNodeID)
```
tiagolobocastro: Should this be a warning?

pkg/lvm/volume.go
```diff
@@ -218,6 +218,9 @@ func RemoveVolFinalizer(vol *apis.LVMVolume) error {
 	vol.Finalizers = nil
 
 	_, err := volbuilder.NewKubeclient().WithNamespace(LvmNamespace).Update(vol)
+	if err != nil {
+		klog.Infof("Finalizer removed successfully for %s", vol.Name)
```
tiagolobocastro: Do we need this as info or is debug fine?

pkg/lvm/volume.go
tiagolobocastro: I guess it means the deletion timestamp is being set?

pkg/driver/controller.go
```diff
@@ -412,6 +434,14 @@ func (cs *controller) deleteVolume(ctx context.Context, volumeID string) error {
 				"failed to handle delete volume request for {%s}", volumeID)
 		}
 	}
+	present := isNodePresent(vol.Spec.OwnerNodeID)
```
tiagolobocastro, line 515: what if the finalizer is already removed? Then we don't need to run this?

tiagolobocastro (commented):
What about existing volumes

pkg/driver/controller.go
abhilashshetty04: finalizer is removed from the node-agent when `lvremove` succeeds. Here we will validate if node is not present. If not, finalizer will not be removed. so will remove finalizer from here.

pkg/driver/controller.go
abhilashshetty04: Yes, will make it warning

pkg/lvm/volume.go
abhilashshetty04: Here we call delete which just marks deletion timestamp as we have finaliser. 

Shall i remove marking deletion timestamp ?

pkg/lvm/volume.go
abhilashshetty04: Will not reach this code if node is present. Would be better to have it in log by default.

tiagolobocastro: Don't we already have the message from above where we call it if the node is not present?
Oh also this check seems reversed, if err != nil then we didn't remove the finalizer?

pkg/lvm/volume.go
tiagolobocastro: I would keep that bit out since it's implied, but fine either way

pkg/driver/controller.go
tiagolobocastro: If the lvm volume has additional finalisers (eg: added by user) but we've already removed ours, on a previous csi call, then we don't need to do this again
We only need to do this stage if our finaliser is present

pkg/driver/controller.go
```diff
@@ -412,6 +434,14 @@ func (cs *controller) deleteVolume(ctx context.Context, volumeID string) error {
...
+	if !present {
+		klog.Infof("Removing finalizer as node %s is not present in cluster", vol.Spec.OwnerNodeID)
+		if err = lvm.RemoveVolFinalizer(vol); err != nil {
+			return nil
```
tiagolobocastro: shouldn't we return err here?

pkg/lvm/volume.go
abhilashshetty04: I will remove it.

pkg/driver/controller.go
abhilashshetty04: Right now this removes all Finalizers.  I am using the function which is called by node-agent once `lvremove` is done.  

https://github.com/openebs/lvm-localpv/blob/develop/pkg/lvm/volume.go#L217C1-L223C2

`lvmvolume` CR is managed by localpv-lvm. Users may not add any finalizers onto it.

pkg/driver/controller.go
abhilashshetty04: Fail the DeleteVolume call? I guess that will be better.

pkg/lvm/volume.go
abhilashshetty04: Yes, err != nil means we fail to remove finalizer, we return err. If not, we move on to check if lvmvolume CR is deleted.

pkg/driver/controller.go
tiagolobocastro: Is this documented somewhere that users may not add finalizers?
It seems we just take the hammer there

pkg/lvm/volume.go
```diff
@@ -164,6 +164,23 @@ func WaitForLVMVolumeDestroy(ctx context.Context, volumeID string) error {
...
+		if err != nil {
+			if k8serror.IsNotFound(err) {
+				return nil
+			}
+			return status.Errorf(codes.Internal,
+				"lvmvolume: destroy wait failed, not able to get the volume %s %s", volumeID, err.Error())
+		}
+		time.Sleep(time.Duration(100 * time.Millisecond))
+	}
+	return status.Errorf(codes.FailedPrecondition,
```
tiagolobocastro: Is this error code right? Shouldn't this be a timeout error?

pkg/driver/controller.go
abhilashshetty04: We remove finaliser here only if node is not present. In the normal flow the finaliser will be removed using `lvm.RemoveVolFinalizer(vol); `by node-agent after doing `lvremove`

 In the previous step we just mark the deletion timestamp by executing delete on the CR.

pkg/lvm/volume.go
abhilashshetty04: I will try the Timeout

abhilashshetty04: I think TimeOut is generated by gRPC framework. Its not available in the library.

var strToCode = map[string]Code{
	`"OK"`: OK,
	`"CANCELLED"`:/* [sic] */ Canceled,
	`"UNKNOWN"`:             Unknown,
	`"INVALID_ARGUMENT"`:    InvalidArgument,
	`"DEADLINE_EXCEEDED"`:   DeadlineExceeded,
	`"NOT_FOUND"`:           NotFound,
	`"ALREADY_EXISTS"`:      AlreadyExists,
	`"PERMISSION_DENIED"`:   PermissionDenied,
	`"RESOURCE_EXHAUSTED"`:  ResourceExhausted,
	`"FAILED_PRECONDITION"`: FailedPrecondition,
	`"ABORTED"`:             Aborted,
	`"OUT_OF_RANGE"`:        OutOfRange,
	`"UNIMPLEMENTED"`:       Unimplemented,
	`"INTERNAL"`:            Internal,
	`"UNAVAILABLE"`:         Unavailable,
	`"DATA_LOSS"`:           DataLoss,
	`"UNAUTHENTICATED"`:     Unauthenticated,
}

tiagolobocastro: `DEADLINE_EXCEEDED`

pkg/lvm/volume.go
abhilashshetty04: @tiagolobocastro , Fixed, This needs to be logged on success. 

It is used by node-plugin also. Seems okay to log that finaliser is removed. wdyt?

tiagolobocastro: There's already the log prior above so probably not needed as info, but up to you

tiagolobocastro (approved):

pkg/lvm/volume.go
abhilashshetty04: Ill remove this. If we see DeleteVolume after prior log then its implied that we couldn't remove Finaliser.

PR #416, opened 2025-10-05
build: make develop branch build again

jochenseeber:
### Why?

- Tried to build develop branch, but did not compile
- Mostly because some dependencies required Go 1.23, but the project was set to use Go 1.19
- PR upgrades to Go 1.24, makes develop build again, fixes unit tests and CI test

### What this PR does?

- Fixes unit tests and CI tests
- Updated Go to 1.24.7
- Required some follow up changes to make everything run again
    - Updated golangci-lint to 2.5.0
    - Updated k8s modules to 0.34.1
    - Updated code generation to write to the correct location
    - Updated Nix configuration
- CI script changes:
    - Added own fio container as there was no arm64 support
    - Added a provider option to select k3s
    - Added a data directory option because tests were failing on `/tmp`
    - Made script run as non-root user with sudo permission
- Updated Nix configuration to use Go 1.24.7
- Fixed Go lint errors
- Added a Vagrantfile to run a development VM for non-Linux users

**Checklist:**
- [x] Fixes #415
- [x] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [x] Has the change log section been updated? → No functional changes
- [x] Commit has unit tests → No new tests, fixes existing ones
- [x] Commit has integration tests → Fixes existing ones

PR #417, opened 2025-10-06
feat(restore): add controller changes to enable restore from thin snapshot

rohan2794:
Why is this PR required?
This PR update new fields into LVM volume and snapshot CRD which is required for restore from snapshot.

**What this PR does?**:
- Add two new fields to the LVM Snapshot CR: **`thinProvision`** and **`lvSize`**.
    - **`thinProvision`**: Indicates whether the snapshot was created from a thin-provisioned volume.  
    - **`lvSize`**: Records the capacity of the snapshot lv at the time the snapshot was taken.
- Add a new field **`source`** to the LVM Volume CR. It defines the data source for the volume.
    - Can be a **snapshot**  
    - Or an **existing volume**
- Update snapshot and volume CRD yaml
- Validate restore from snapshot request.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/deploy/lvm-operator.yaml
+++ b/deploy/lvm-operator.yaml
@@ -114,6 +114,13 @@ spec:
                 - "yes"
                 - "no"
                 type: string
+              source:
+                description: |-
+                  Source defines the data source from which the volume should be created.
+                  It can reference either a snapshot or an existing volume.
+                  If not specified, a standard LVM volume will be created.
+                  If specified, the new volume will be created using the defined data source.
+                type: string
               thinProvision:
                 description: |-
                   ThinProvision specifies whether logical volumes can be thinly provisioned.
@@ -236,6 +243,10 @@ spec:
               snapSize:
                 description: SnapSize specifies the space reserved for the snapshot
                 type: string
+              thinProvision:
+                description: ThinProvision specifies whether the snapshot is thin-provisioned
+                  or not.
+                type: boolean
               volGroup:
                 description: VolGroup specifies the name of the volume group where
                   the snapshot has been created.
@@ -248,6 +259,14 @@ spec:
             description: SnapStatus string that reflects if the snapshot was created
               successfully
             properties:
+              lvSize:
+                anyOf:
+                - type: integer
+                - type: string
+                description: LvSize specifies the size allocated for the snapshot
+                  in the VG
+                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                x-kubernetes-int-or-string: true
               state:
                 type: string
             type: object
--- a/deploy/yamls/lvmsnapshot-crd.yaml
+++ b/deploy/yamls/lvmsnapshot-crd.yaml
@@ -60,6 +60,10 @@ spec:
               snapSize:
                 description: SnapSize specifies the space reserved for the snapshot
                 type: string
+              thinProvision:
+                description: ThinProvision specifies whether the snapshot is thin-provisioned
+                  or not.
+                type: boolean
               volGroup:
                 description: VolGroup specifies the name of the volume group where
                   the snapshot has been created.
@@ -72,6 +76,14 @@ spec:
             description: SnapStatus string that reflects if the snapshot was created
               successfully
             properties:
+              lvSize:
+                anyOf:
+                - type: integer
+                - type: string
+                description: LvSize specifies the size allocated for the snapshot
+                  in the VG
+                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                x-kubernetes-int-or-string: true
               state:
                 type: string
             type: object
--- a/pkg/apis/openebs.io/lvm/v1alpha1/lvmsnapshot.go
+++ b/pkg/apis/openebs.io/lvm/v1alpha1/lvmsnapshot.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -47,6 +48,9 @@ type LVMSnapshotSpec struct {
 
 	// SnapSize specifies the space reserved for the snapshot
 	SnapSize string `json:"snapSize,omitempty"`
+
+	// ThinProvision specifies whether the snapshot is thin-provisioned or not.
+	ThinProvision bool `json:"thinProvision,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -63,4 +67,7 @@ type LVMSnapshotList struct {
 // SnapStatus string that reflects if the snapshot was created successfully
 type SnapStatus struct {
 	State string `json:"state,omitempty"`
+
+	// LvSize specifies the size allocated for the snapshot in the VG
+	LvSize resource.Quantity `json:"lvSize,omitempty"`
 }
--- a/pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
+++ b/pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
@@ -85,6 +85,12 @@ type VolumeInfo struct {
 	// thinProvision i.e. logical volumes that are larger than the available extents.
 	// +kubebuilder:validation:Enum=yes;no
 	ThinProvision string `json:"thinProvision,omitempty"`
+
+	// Source defines the data source from which the volume should be created.
+	// It can reference either a snapshot or an existing volume.
+	// If not specified, a standard LVM volume will be created.
+	// If specified, the new volume will be created using the defined data source.
+	Source string `json:"source,omitempty"`
 }
 
 // VolStatus string that specifies the current state of the volume provisioning request.
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -22,7 +22,7 @@ import (
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/klog/v2"
+	klog "k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 
 	lvmapi "github.com/openebs/lvm-localpv/pkg/apis/openebs.io/lvm/v1alpha1"
@@ -243,6 +243,42 @@ func (cs *controller) init() error {
 	return nil
 }
 
+func validateRestoreRequestFromSnapshot(lvmSnapCr *lvmapi.LVMSnapshot,
+	req *csi.CreateVolumeRequest,
+	params *VolumeParams,
+) error {
+	// verify that the snapshot is of thin type
+	// only thin snapshot can be used to restore a volume
+	// also verify that the new volume to be created from snapshot
+	// is also thin provisioned
+	if !lvmSnapCr.Spec.ThinProvision {
+		return status.Errorf(codes.InvalidArgument, "snapshot %s is not of thin type", lvmSnapCr.Name)
+	} else if params.ThinProvision != lvm.YES {
+		return status.Errorf(codes.InvalidArgument, "only thin restores supported today.")
+	}
+
+	vgPattern := fmt.Sprintf("^%v$", lvmSnapCr.Spec.VolGroup)
+
+	// verify that the volume group of the new volume to be created
+	// from snapshot is same as that of the snapshot
+	if params.VgPattern.String() != vgPattern {
+		return status.Errorf(codes.InvalidArgument, "volume group of the restore volume can't be different from that of source (snapshot %s)",
+			lvmSnapCr.Name)
+	}
+
+	reqCapacity := getRoundedCapacity(
+		req.GetCapacityRange().RequiredBytes)
+
+	// verify that the capacity of the restore volume to be created
+	// from snapshot should equal to the snapshot LV size
+	if lvmSnapCr.Status.LvSize.Value() != reqCapacity {
+		return status.Errorf(codes.OutOfRange, "capacity of the restore volume %v to be created from snapshot must be equal to snapshot LV size %d",
+			reqCapacity,
+			lvmSnapCr.Status.LvSize.Value())
+	}
+	return nil
+}
+
 // CreateLVMVolume create new lvm volume for csi volume request
 func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
 	params *VolumeParams) (*lvmapi.LVMVolume, error) {
@@ -278,22 +314,50 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
 			}
 		}
 	}
+	var owner string             // owner node where volume is to be created
+	var lvmSnapshotCrName string // snapshot name if volume is to be created from snapshot
+	var volGroup string          // volume group where volume is to be created
+	// if volume is to be created from snapshot, verify request with snapshot values
+	if req.GetVolumeContentSource() != nil && req.GetVolumeContentSource().GetSnapshot() != nil {
+		snapId := req.GetVolumeContentSource().GetSnapshot().GetSnapshotId()
+		// snapshot ID is of the form <volume-name>@<snapshot-name>
+		parts := strings.Split(snapId, "@")
+		if len(parts) != 2 {
+			return nil, status.Errorf(codes.Internal,
+				"invalid snapshot ID %s", snapId)
+		}
+		lvmSnapCr, err := lvm.GetLVMSnapshot(parts[1])
+		if err != nil {
+			return nil, status.Errorf(codes.NotFound, "snapshot %s not found: %s", parts[1], err.Error())
+		}
+		// verify snapshot values like thinSnapshot,owner node, thinprovision and volume group and capacity for the new volume
+		if err = validateRestoreRequestFromSnapshot(lvmSnapCr, req, params); err != nil {
+			return nil, err
+		}
 
-	nmap, err := getNodeMap(params.Scheduler, params.VgPattern)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "get node map failed : %s", err.Error())
+		owner = lvmSnapCr.Spec.OwnerNodeID
+		lvmSnapshotCrName = lvmSnapCr.Name
+		volGroup = lvmSnapCr.Spec.VolGroup
 	}
 
-	// run the scheduler
-	selected := schd.Scheduler(req, nmap)
+	// if owner is not already set from snapshot, run the scheduler
+	// to select the owner node for the volume
+	if owner == "" {
+		nmap, err := getNodeMap(params.Scheduler, params.VgPattern)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "get node map failed : %s", err.Error())
+		}
 
-	if len(selected) == 0 {
-		return nil, status.Error(codes.Internal, "scheduler failed, not able to select a node to create the PV")
+		// run the scheduler
+		selected := schd.Scheduler(req, nmap)
+		if len(selected) == 0 {
+			return nil, status.Error(codes.Internal, "scheduler failed, not able to select a node to create the PV")
+		}
+
+		owner = selected[0]
+		klog.Infof("scheduling the volume %s/%s on node %s",
+			params.VgPattern.String(), volName, owner)
 	}
-
-	owner := selected[0]
-	klog.Infof("scheduling the volume %s/%s on node %s",
-		params.VgPattern.String(), volName, owner)
 
 	volObj, err := volbuilder.NewBuilder().
 		WithName(volName).
@@ -305,7 +369,20 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
 		WithThinProvision(params.ThinProvision).Build()
 
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return nil, status.Errorf(codes.Internal,
+			"failed to build lvm volume object for %s: {%s}",
+			volName, err.Error())
+	}
+
+	if lvmSnapshotCrName != "" {
+		volObj, err = volbuilder.BuildFrom(volObj).
+			WithSource(lvmSnapshotCrName).
+			WithVolGroup(volGroup).Build()
+		if err != nil {
+			return nil, status.Errorf(codes.Internal,
+				"failed to update lvm volume object with source and volgroup for %s: {%s}",
+				volName, err.Error())
+		}
 	}
 
 	vol, err = lvm.ProvisionVolume(volObj)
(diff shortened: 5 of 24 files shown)


deploy/lvm-operator.yaml
```diff
@@ -230,6 +237,18 @@ spec:
               snapSize:
                 description: SnapSize specifies the space reserved for the snapshot
                 type: string
+              sourceVolumeCapacity:
```
dsharma-dc: The term `source` (origin) here and the actual `source` field above are logically different, so I'm wondering if we should call this one something like `snapOriginVolumeCapacity` or such.

deploy/lvm-operator.yaml
```diff
@@ -230,6 +237,18 @@ spec:
...
+                  volume at the time the snapshot was created.
+                type: string
+              thinSnapshot:
+                description: ThinSnapshot indicates whether the snapshot should be
+                  thin-provisioned. When set to "yes", the LVM LocalPV driver creates
+                  a thin snapshot of the logical volume.
+                enum:
+                - "yes"
+                - "no"
+                type: string
```
dsharma-dc: Isn't this more like a readonly field for now? and not a user instruction. This we set after taking a thin snapshot of the origin.

deploy/yamls/lvmsnapshot-crd.yaml
```diff
@@ -56,6 +56,18 @@ spec:
               snapSize:
                 description: SnapSize specifies the space reserved for the snapshot
                 type: string
+              sourceVolumeCapacity:
```
dsharma-dc: Similarly here:
snapOriginVolumeCapacity

pkg/apis/openebs.io/lvm/v1alpha1/lvmsnapshot.go
```diff
@@ -49,6 +49,18 @@ type LVMSnapshotSpec struct {
 	// SnapSize specifies the space reserved for the snapshot
 	// +kubebuilder:validation:Required
 	SnapSize string `json:"snapSize,omitempty"`
+
+	// SourceVolumeCapacity represents the capacity of the source volume
```
dsharma-dc: In general, use the word `origin` to refer to this volume.

```suggestion
	// SourceVolumeCapacity represents the capacity of the origin volume
```

pkg/driver/controller.go
```diff
@@ -278,24 +320,57 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
...
+		if len(parts) != 2 {
+			return nil, status.Errorf(codes.Internal,
+				"invalid snapshot ID %s", snapId)
+		}
+		lvmSnapCr, err := lvm.GetLVMSnapshot(parts[1])
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "snapshot %s not found: %s", parts[1], err.Error())
+		}
+		// verify snapshot values like thinSnapshot,owner node, thinprovision and volume group and capacity for the new volume
+		if err = validateRestoreRequestFromSnapshot(parts[1], req, params); err != nil {
```
dsharma-dc: can we pass `lvmSnapCr` here since you already do `GetLVMSnapshot` again inside?

pkg/driver/controller.go
```diff
@@ -243,6 +243,48 @@ func (cs *controller) init() error {
...
+	}
+
+	// verify that the snapshot is of thin type
+	// only thin snapshot can be used to restore a volume
+	// also verify that the new volume to be created from snapshot
+	// is also thin provisioned
+	if lvmSnapCr.Spec.ThinSnapshot != lvm.YES {
+		return status.Errorf(codes.Internal, "snapshot %s is not of thin type", lvmSnapshotCrName)
+	} else if params.ThinProvision != lvm.YES {
+		return status.Errorf(codes.Internal, "volume from snapshot %s must be thin provisioned", lvmSnapshotCrName)
```
dsharma-dc: May be say that we support only thin restores only today.

pkg/driver/controller.go
```diff
@@ -243,6 +243,48 @@ func (cs *controller) init() error {
...
+	}
+
+	vgPattern := fmt.Sprintf("^%v$", lvmSnapCr.Spec.VolGroup)
+
+	// verify that the volume group of the new volume to be created
+	// from snapshot is same as that of the snapshot
+	if params.VgPattern.String() != vgPattern {
+		return status.Errorf(codes.Internal, "volume group/ vg pattern of the new volume to be created from snapshot %s must be same as that of the snapshot",
```
dsharma-dc:
```suggestion
		return status.Errorf(codes.Internal, "volume group of the restore volume can't be different from that of source (snapshot %s)",
```

pkg/driver/controller.go
```diff
@@ -243,6 +243,48 @@ func (cs *controller) init() error {
...
+			lvmSnapshotCrName)
+	}
+
+	reqCapacity := strconv.FormatInt(getRoundedCapacity(
+		req.GetCapacityRange().RequiredBytes), 10)
+
+	// verify that the capacity of the new volume to be created
+	// from snapshot should equal to the source volume capacity at the time of snapshot creation
```
dsharma-dc: This might need to change in future where we could automatically cater to larger sizes by doing extend at the time of restore.

```suggestion
	// from snapshot should equal to the origin volume capacity at the time of snapshot creation
```

deploy/lvm-operator.yaml
rohan2794: ok, i will update the CRD field to snapOriginVolumeCapacity .

pkg/driver/controller.go
rohan2794: yes, it might change when larger restore volume capacity is supported.

deploy/lvm-operator.yaml
rohan2794: yes, thinSnapshot is readonly field.
I have updated the description.

pkg/driver/controller.go
```diff
@@ -278,24 +314,57 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
...
+			return nil, status.Errorf(codes.Internal, "get node map failed : %s", err.Error())
+		}
+
+		// run the scheduler
+		selected := schd.Scheduler(req, nmap)
 
-	if len(selected) == 0 {
-		return nil, status.Error(codes.Internal, "scheduler failed, not able to select a node to create the PV")
-	}
+		if len(selected) == 0 {
```
dsharma-dc, line 353: Is there an indentation issue here? github shows tabs

dsharma-dc (approved):
I don't remember if we discussed  this - can the restored LV be independently expanded without any issue?

Also, please add a couple of CI test here or in a new PR.

deploy/yamls/lvmsnapshot-crd.yaml
```diff
@@ -53,9 +53,17 @@ spec:
                   can not be edited after the snapshot has been provisioned.
                 minLength: 1
                 type: string
+              snapOriginVolumeCapacity:
+                description: SnapOriginVolumeCapacity represents the capacity of the
+                  origin volume at the time the snapshot was created.
+                type: string
```
tiagolobocastro: Can we modify code to generate this?

```suggestion
                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                type: string
                x-kubernetes-int-or-string: true
```

tiagolobocastro: Btw do we need this on the spec?
Also is there a race in case volume gets resized before snapshot is created?

pkg/apis/openebs.io/lvm/v1alpha1/lvmvolume.go
```diff
@@ -87,6 +87,13 @@ type VolumeInfo struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=yes;no
 	ThinProvision string `json:"thinProvision,omitempty"`
+
+	// Source defines the data source from which the volume should be created.
+	// It can reference either a snapshot or an existing volume.
+	// If not specified, a standard LVM volume will be created.
+	// If specified, the new volume will be created using the defined data source.
+	// +kubebuilder:validation:Required
```
tiagolobocastro: This should be optional

```suggestion
```

pkg/collector/lv_collector.go
```diff
@@ -18,7 +18,7 @@ package collector
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
-	"k8s.io/klog/v2"
+	klog "k8s.io/klog/v2"
```
tiagolobocastro, line 21: All these seem to be done by #416 already

pkg/driver/controller.go
```diff
@@ -243,6 +243,42 @@ func (cs *controller) init() error {
...
 
+func validateRestoreRequestFromSnapshot(lvmSnapCr *lvmapi.LVMSnapshot,
+	req *csi.CreateVolumeRequest,
+	params *VolumeParams,
+) error {
+	// verify that the snapshot is of thin type
+	// only thin snapshot can be used to restore a volume
+	// also verify that the new volume to be created from snapshot
+	// is also thin provisioned
+	if !lvmSnapCr.Spec.Thin {
```
tiagolobocastro: It may be worth considering using `ThinProvision` string style to keep in line with existing usage

pkg/driver/controller.go
```diff
@@ -243,6 +243,42 @@ func (cs *controller) init() error {
...
+		return status.Errorf(codes.Internal, "snapshot %s is not of thin type", lvmSnapCr.Name)
+	} else if params.ThinProvision != lvm.YES {
+		return status.Errorf(codes.Internal, "only thin restores supported today.")
```
tiagolobocastro: Shouldn't these be `INVALID_ARGUMENT` ? See the [spec](https://github.com/container-storage-interface/spec/blob/e6fc13ea4d529db12e211ef79c924ee3186c39d5/spec.md?plain=1#L1262)
 
```suggestion
		return status.Errorf(codes.InvalidArgument, "only thin restores supported today")
```

pkg/driver/controller.go
```diff
@@ -278,24 +314,57 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
...
+		snapId := req.GetVolumeContentSource().GetSnapshot().GetSnapshotId()
+		// snapshot ID is of the form <volume-name>@<snapshot-name>
+		parts := strings.Split(snapId, "@")
+		if len(parts) != 2 {
+			return nil, status.Errorf(codes.Internal,
+				"invalid snapshot ID %s", snapId)
+		}
+		lvmSnapCr, err := lvm.GetLVMSnapshot(parts[1])
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "snapshot %s not found: %s", parts[1], err.Error())
```
tiagolobocastro: Again as per csi, this should be NotFound error?

```suggestion
			return nil, status.Errorf(codes.NotFound, "snapshot %s not found: %s", parts[1], err.Error())
```

pkg/driver/controller.go
```diff
@@ -278,24 +314,57 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
 			}
 		}
 	}
+	var volObj *lvmapi.LVMVolume
```
tiagolobocastro: should we declare this further below closer to when it's used? I'm fuzzy on the go idiomatics

pkg/driver/controller.go
```diff
@@ -278,24 +314,57 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
...
+		}
 
-	nmap, err := getNodeMap(params.Scheduler, params.VgPattern)
+		owner = lvmSnapCr.Spec.OwnerNodeID
+		lvmSnapshotCrName = lvmSnapCr.Name
+		volGroup = lvmSnapCr.Spec.VolGroup
+	}
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "get node map failed : %s", err.Error())
+		return nil, status.Error(codes.Internal, err.Error())
```
tiagolobocastro: there's no error to check here now? As you've moved the call to `getNodeMap` below?

pkg/driver/controller.go
```diff
@@ -305,7 +374,20 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
...
+			volName, err.Error())
+	}
+
+	if lvmSnapshotCrName != "" {
+		volObj, err = volbuilder.BuildFrom(volObj).
+			WithSource(lvmSnapshotCrName).
+			WithVolGroup(volGroup).Build()
+		if err != nil {
+			return nil, status.Errorf(codes.Internal,
+				"failed to build lvm volume object for %s: {%s}",
```
tiagolobocastro: might want to modify slightly from the one above so we can tell apart

pkg/driver/controller.go
```diff
@@ -337,9 +419,8 @@ func (cs *controller) CreateVolume(
 	contentSource := req.GetVolumeContentSource()
 
 	var vol *lvmapi.LVMVolume
-	if contentSource != nil && contentSource.GetSnapshot() != nil {
-		return nil, status.Error(codes.Unimplemented, "")
-	} else if contentSource != nil && contentSource.GetVolume() != nil {
+	if contentSource != nil && contentSource.GetVolume() != nil {
+		_ = vol // bypass unused variable warning
```
tiagolobocastro: can't we just remove the else clause since we're returning here?

tiagolobocastro (commented):
Can you also add some restore tests to https://github.com/openebs/lvm-localpv/tree/develop/tests?

deploy/yamls/lvmsnapshot-crd.yaml
abhilashshetty04: Shall we have it as int64?

pkg/driver/controller.go
abhilashshetty04: `InvalidArgument` if Volume is a content source? We return `Unimplemented` now.

pkg/driver/controller.go
```diff
@@ -243,6 +243,42 @@ func (cs *controller) init() error {
...
+			lvmSnapCr.Name)
+	}
+
+	reqCapacity := strconv.FormatInt(getRoundedCapacity(
+		req.GetCapacityRange().RequiredBytes), 10)
+
+	// verify that the capacity of the new volume to be created
+	// from snapshot should equal to the origin volume capacity at the time of snapshot creation
+	if lvmSnapCr.Spec.SnapOriginVolumeCapacity != reqCapacity {
+		return status.Errorf(codes.Internal, "capacity of the new volume %s to be created from snapshot must be equal to the source volume capacity %s",
```
abhilashshetty04: OutOfRange here.. https://github.com/container-storage-interface/spec/blob/e6fc13ea4d529db12e211ef79c924ee3186c39d5/spec.md?plain=1#L1266

abhilashshetty04: This doesn't seem right.  If source volume is resized after taking snapshot?

abhilashshetty04: Can we reference by saying: SnapOriginVolumeCapacity indicated in lvmsnapshot CR.

deploy/yamls/lvmsnapshot-crd.yaml
tiagolobocastro: We can have it as [Quantity](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/), like [here](https://github.com/openebs/lvm-localpv/blob/f103ba5e3fa2521a3bc3e6e9c53608942b950b5d/pkg/apis/openebs.io/lvm/v1alpha1/lvmnode.go#L55).

pkg/driver/controller.go
tiagolobocastro: Yes that's right. 
`Unimplemented` is likely not correct, as that's for `RPC handler` not implemented.

pkg/driver/controller.go
tiagolobocastro: This brings another point, what if volume is resized but without being mounted. And you take snapshot at this point. Now filesystem will be missing a resize in the snapshot? So does this mean that we do need to implement filesystem resize on snapshot creation/first use?

abhilashshetty04: When volume resized before mounting. We update lvmvolume cr but don't do `lvextend`. If we create a snapshot and restore from that volume it will consider resized capacity. Parent volume (lv) will have older capacity until it gets mounted. This should not be a problem i believe. Parent lv will get resized when it gets mounted  cc: @rohan2794 , @dsharma-dc

deploy/yamls/lvmsnapshot-crd.yaml
rohan2794: Updated `snapOriginVolumeCapacity ` type to [Quantity](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/)

pkg/collector/lv_collector.go
rohan2794: yes, lint was failing for me as well. Rebase will resolve this .

pkg/driver/controller.go
rohan2794: updated to `ThinProvision `

pkg/driver/controller.go
rohan2794: changed it to `InvalidArgument `

pkg/driver/controller.go
rohan2794: fixed

pkg/driver/controller.go
rohan2794: fixed

pkg/driver/controller.go
rohan2794: updated to `OutOfRange `

rohan2794:
> Can you also add some restore tests to https://github.com/openebs/lvm-localpv/tree/develop/tests?

I will add CI test in different PR .

rohan2794:
> I don't remember if we discussed this - can the restored LV be independently expanded without any issue?
> 
> Also, please add a couple of CI test here or in a new PR.

Yes, restored thin LV can be independently expanded

abhilashshetty04 (approved):

abhilashshetty04:
@rohan2794 , Please add ci tests also.

rohan2794:
> @rohan2794 , Please add ci tests also.

I will add CI test in different PR once node-plugin changes are merged.

pkg/driver/controller.go
```diff
@@ -243,6 +244,42 @@ func (cs *controller) init() error {
...
+func validateRestoreRequestFromSnapshot(lvmSnapCr *lvmapi.LVMSnapshot,
+	req *csi.CreateVolumeRequest,
+	params *VolumeParams,
+) error {
+	// verify that the snapshot is of thin type
+	// only thin snapshot can be used to restore a volume
+	// also verify that the new volume to be created from snapshot
+	// is also thin provisioned
+	if !lvmSnapCr.Spec.ThinProvision {
+		return status.Errorf(codes.Internal, "snapshot %s is not of thin type", lvmSnapCr.Name)
```
tiagolobocastro: This one also?

```suggestion
		return status.Errorf(codes.InvalidArgument, "snapshot %s is not of thin type", lvmSnapCr.Name)
```

pkg/driver/controller.go
```diff
@@ -243,6 +244,42 @@ func (cs *controller) init() error {
...
+	} else if params.ThinProvision != lvm.YES {
+		return status.Errorf(codes.InvalidArgument, "only thin restores supported today.")
+	}
+
+	vgPattern := fmt.Sprintf("^%v$", lvmSnapCr.Spec.VolGroup)
+
+	// verify that the volume group of the new volume to be created
+	// from snapshot is same as that of the snapshot
+	if params.VgPattern.String() != vgPattern {
+		return status.Errorf(codes.Internal, "volume group of the restore volume can't be different from that of source (snapshot %s)",
```
tiagolobocastro: This one also?

```suggestion
		return status.Errorf(codes.InvalidArgument, "volume group of the restore volume can't be different from that of source (snapshot %s)",
```

pkg/driver/controller.go
```diff
@@ -337,22 +415,19 @@ func (cs *controller) CreateVolume(
 	contentSource := req.GetVolumeContentSource()
 
 	var vol *lvmapi.LVMVolume
-	if contentSource != nil && contentSource.GetSnapshot() != nil {
+	if contentSource != nil && contentSource.GetVolume() != nil {
+		_ = vol // bypass unused variable lint warning
```
tiagolobocastro: if you declare variable close to its use you won't need this bypass

pkg/driver/controller.go
tiagolobocastro: But the cloned volume will not get resized?

tiagolobocastro (approved):

pkg/driver/controller.go
rohan2794: fixed

pkg/driver/controller.go
rohan2794: fixed

pkg/driver/controller.go
rohan2794: make sense, fixed

pkg/driver/controller.go
abhilashshetty04: No, it won't. Here there is one more thing to consider. Even though the lvmsnapshot, lvmvolume will reference the newer capacity which is pending lvextend. The snapshot and restore will have the parent volume size. As lvcreate of snapshot and restore reference a LV.

Resize On unmounted volume. Considering a thin pvc being resized from 1GiB to 2GiB.

`parent unmounted` (PVC spec.size: 2GiB, PVC state.size 1GiB, CR: 2GiB, lv size 1GiB) 
`lvmsnapshot` (lvmsnapshot CR 2GiB, lvsize 1GiB) 
`cloned volume `(PVC size: 2GiB, PVC state.size: ? , CR 2GiB, lvsize 1GiB)

`parent mounted `(PVC spec.size: 2GiB, PVC state.size 2GiB, CR 2GiB, lvsize 2GiB)
`lvmsnapshot` (CR 2GiB, lvsize 1GiB)
`cloned volume` (PVC state.size (not sure) , CR 2GiB,  lvsize 1GiB) -> This lv will not be resized after original volume gets lvextend post mount. 

However, We might need to fix the discrepancy by having actual lvsize in CR state perhaps.

rohan2794: With [this commit](https://github.com/openebs/lvm-localpv/pull/417/commits/b9cae07a2fc46402d0cc5c49ac1f2cc3202dfa44)
, I removed the `SnapOriginVolumeCapacity` field from the Snapshot CR spec and added the snapshot LV size (LvSize) field under the Snapshot CR status.

The LvSize field is populated by the node plugin once the snapshot is created. During a restore request, the controller uses `LvSize` value to validate the requested capacity against the actual snapshot size. This ensures consistency and eliminates any potential capacity discrepancies.

deploy/yamls/lvmsnapshot-crd.yaml
```diff
@@ -72,8 +76,18 @@ spec:
...
+                - type: integer
+                - type: string
+                description: LvSize specifies the size allocated for the snapshot
+                  in the VG
+                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                x-kubernetes-int-or-string: true
               state:
                 type: string
+            required:
+            - lvSize
```
tiagolobocastro: this is a breaking change, should not be made a required parameter

pkg/apis/openebs.io/lvm/v1alpha1/lvmsnapshot.go
```diff
@@ -63,4 +67,8 @@ type LVMSnapshotList struct {
 // SnapStatus string that reflects if the snapshot was created successfully
 type SnapStatus struct {
 	State string `json:"state,omitempty"`
+
+	// LvSize specifies the size allocated for the snapshot in the VG
+	// +kubebuilder:validation:Required
```
tiagolobocastro: This is why it's being set as required

```suggestion
```

deploy/yamls/lvmsnapshot-crd.yaml
rohan2794: made LvSize optional , [fixed](https://github.com/openebs/lvm-localpv/pull/417/files/0dcbacd19329e6fa3336ad958c30ab1317baa928..e524578c3794418c7393269c269aff26ba512d36)

abhilashshetty04:
@rohan2794 LGTM. Please squash the commits

PR #424, opened 2025-11-13
fix: update snapshot create response with restore size for snapshot

rohan2794:
This PR update create snapshot response :

- In case of thick volume snapshot restore size to 0
- In case of thin snapshot, wait for lvm snapshot CR to be ready and set snapshot restore size to snapshot lv size.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -365,7 +365,6 @@ func CreateLVMVolume(ctx context.Context, req *csi.CreateVolumeRequest,
 	}
 	klog.Infof("scheduling the volume %s/%s on node %s",
 		params.VgPattern.String(), volName, owner)
-		
 	volObj, err := volbuilder.NewBuilder().
 		WithName(volName).
 		WithCapacity(capacity).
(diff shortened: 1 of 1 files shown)


abhilashshetty04:
@rohan2794 , Shall we change PR title as there is no change to ReadyToUse field?

rohan2794:
> @rohan2794 , Shall we change PR title as there is no change to ReadyToUse field?

Yes, make sense

pkg/driver/controller.go
```diff
@@ -751,14 +771,55 @@ func (cs *controller) CreateSnapshot(
...
+			WithSnapshotID(req.SourceVolumeId+"@"+req.Name).
+			WithCreationTime(snapTimeStamp, 0).
+			WithReadyToUse(state == lvm.LVMStatusReady).
+			WithSize(restoreSize).
+			Build(),
+		nil
+
+}
+
+// waitForLVMSnapshotProcessed waits for completion of any processing of lvm snapshot.
```
abhilashshetty04: It returns restoreSize or err.

rohan2794: fixed

abhilashshetty04 (approved):

pkg/driver/controller.go
```diff
@@ -751,16 +755,45 @@ func (cs *controller) CreateSnapshot(
...
+		snap, err = lvm.GetLVMSnapshot(snapName)
+		if err == nil {
+			if snap.Status.State == lvm.LVMStatusReady {
+				break
+			}
+		}
+		time.Sleep(time.Duration(100 * time.Millisecond))
+	}
+	return snap, err
+}
```
Abhinandan-Purkait, line 797: Why do we need this here? Can we just not fail anyway and let the retry handle it, which will anyhow happen here once 1 second passes? Also isn't 1s too low?

rohan2794: @Abhinandan-Purkait  Seems not correct to wait at all after creating LVM snapshot CR , this might lead to problem where user try to create snapshot at scale as `external-snapshotter `will have to queue all the snapshot request with exponential backoff.
I will make wait timeout to 2 seconds with 100 millisecond retry interval.

rohan2794: [fixed](https://github.com/openebs/lvm-localpv/pull/424/files/37546b0f84032d2a26dce51f39d0e2fa5dd116c3..aba7b1a71fb7801e1888bc2fcdb53f6f9c76fb07)

Abhinandan-Purkait (approved):

PR #436, opened 2026-01-23
fix(provisioning): Add ownerRef to provisioned volumes

Gufran:
## Pull Request template

Currently the volumes hang around when the node is deleted, which leads to operational toil and alerts. Since the node and the csinode are both gone, it is safe to assume that the actual volume is also lost, in that case we can rely on the built in mechanism to clear the stale references. This is how local-static-provisioner does it as well.

**What this PR does?**:

Add ownerRef to volumes

no

none

none

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests


abhilashshetty04:
Hi @Gufran , We have added fix sometime back: https://github.com/openebs/lvm-localpv/pull/413

Basically we cleanup `lvmvolume` CR if we can't find the k8s node in the cluster.

Gufran:
thanks for pointing it out. I'll update our version and see if that helps. any specific reason to not use owner references?

tiagolobocastro:
> thanks for pointing it out. I'll update our version and see if that helps. any specific reason to not use owner references?

One downside could be user would have to recreate the CRs if they also move the disk to another node?

What about the lvmnode CR btw?

I'm probably ok with getting this in, but not enabled by default

Gufran:
> One downside could be user would have to recreate the CRs if they also move the disk to another node?

This is much more likely to happen in a cloud infra compared to what I deal with on a daily basis. Didn't occur to me to think about that!


> What about the lvmnode CR btw?
> I'm probably ok with getting this in, but not enabled by default

I can update lvmnode as well and put this change behind a config that is disabled by default. Will update the PR soon.

abhilashshetty04:
> thanks for pointing it out. I'll update our version and see if that helps. any specific reason to not use owner references?

Fix which i pointed out will work if user deletes PVC whose lvmvolumes are in nodes which are not in the cluster anymore. I think we cleaning up backend volumes when PVC still exist does not seem right.

tiagolobocastro:
> > One downside could be user would have to recreate the CRs if they also move the disk to another node?
> 
> This is much more likely to happen in a cloud infra compared to what I deal with on a daily basis. Didn't occur to me to think about that!
> 
> > What about the lvmnode CR btw?
> > I'm probably ok with getting this in, but not enabled by default
> 
> I can update lvmnode as well and put this change behind a config that is disabled by default. Will update the PR soon.

@Gufran any updates here, on whether the current approach we have is sufficient to address your needs?

abhilashshetty04:
Hi @Gufran , Closing this due to lack of progress. Feel free to reopen if you want to take it forward.

PR #450, opened 2026-02-11
ci: add trivy scan

pchandra19:
This PR introduces automated security scanning of rendered Helm manifests using Trivy (config mode) in CI.

**What this adds**

- Renders Helm templates during CI
- Scans the rendered manifests for misconfigurations using:
- Severity: HIGH, CRITICAL
- SARIF output format
- Uploads results to GitHub Code Scanning
- Categorizes results under trivy-helm

### DIFF
--- a/.github/workflows/trivy-ci.yml
+++ b/.github/workflows/trivy-ci.yml
@@ -1,0 +1,51 @@
+name: Trivy Check CI
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: '0 3 1,15 * *'
+
+permissions:
+  security-events: write
+  actions: read
+  contents: read
+
+jobs:
+  trivy-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/setup-nix
+        with:
+          preFetchNix: shell.nix
+
+      - name: Helm dependency update
+        run: |
+          nix-shell --run "
+          cd deploy/helm/charts
+          helm dependency update .
+          "
+
+      - name: Render Helm templates
+        run: |
+          nix-shell --run '
+          cd deploy/helm/charts
+          CHART_NAME=$(yq ".name" Chart.yaml)
+          RELEASE_NAME=${RELEASE_NAME:-$CHART_NAME}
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,53 @@
...
+              --format sarif \
+              --output trivy-helm.sarif \
+              rendered.yaml
+          "
+          
+      - name: Upload Helm SARIF
+        uses: github/codeql-action/upload-sarif@v4
+        with:
+          sarif_file: deploy/helm/charts/trivy-helm.sarif
+          category: trivy-helm
```
tiagolobocastro: newline

shell.nix
```diff
@@ -12,6 +12,7 @@ pkgs.mkShell {
     golint
     kubectl
     kubernetes-helm
+    trivy
```
tiagolobocastro: trivy is a large relatively download, could probably skip it here and do `nix-shell -p trivy --run`, wdyt?

.github/workflows/trivy-ci.yml
```diff
@@ -0,0 +1,53 @@
+name: Trivy Check CI
+on:
+  workflow_call: {}
```
tiagolobocastro: How will this run typically?

tiagolobocastro (commented):
nit: should probably be `ci: ` rather than `chore(ci)`

.github/workflows/trivy-ci.yml
pchandra19: As discussed, modified to run on schedule. Twice a month on 1st and 15th at 3:00

shell.nix
pchandra19: Modified it to use aquasecurity/trivy-action@v0.20.0

.github/workflows/trivy-ci.yml
pchandra19: fixed

tiagolobocastro (approved):
What about release branches/tags, we'd have to run those manually?

pchandra19:
> What about release branches/tags, we'd have to run those manually?

Ideally , we wouldn't want to run scanners on release/tags. It'll be too late to fix stuff. But yeah with the way we are setup we can run them manually whenever we want.

PR #451, opened 2026-02-11
fix: update unit test code coverage action

rohan2794:
Unit test code coverage action is failing with unknown option file in v5 version.

**What this PR does?**:
Replace code coverage action `file` parameter with supported `files` parameter.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,6 @@ jobs:
         if: inputs.coverage
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          file: ./coverage.txt
           directory: ./coverage
           flags: unittests
           fail_ci_if_error: true


.github/workflows/tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         if: inputs.coverage
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          file: ./coverage.txt
+          files: ./coverage.txt
```
tiagolobocastro: do we even need to specify files if we already have directory below? Then would work even if we modify coverage to have multiple files?

tiagolobocastro: In fact, it doesn't find any coverate.txt, see: `Some files were not found --- {"not_found_files": ["coverage.txt"]}`

rohan2794: removed files as directory is already there.

tiagolobocastro (approved):

PR #456, opened 2026-02-20
ci: enable mergify for merge/backport

tiagolobocastro:

Enables mergify for merging PRs and also backporting.
Allows us to more easily merge PRs without having to manually rebase them and also allows
for easy backporting.

**What this PR does?**:

We've already been enabling mergify on other repos.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests
- [ ] (Optional) If documentation changes are required, which issue on https://github.com/openebs/website is used to track them:


.github/workflows/pull_request.yml
```diff
@@ -5,50 +5,32 @@ on:
     branches:
       - develop
       - "release/**"
+    paths-ignore:
```
tiagolobocastro: @rohan2794 can we do this here, like on zfs?
Or would this cause issues on merge protection?

rohan2794: @tiagolobocastro paths-ignore will not work if changes are only in path-ignore files due to below reasons: 
- If repository has branch protection enabled (e.g., requiring "lint", "tests", or "csi-driver" jobs to pass before merging), GitHub expects these jobs to report a status (success, failure, or neutral).

- When paths-ignore skips the workflow, no status is reported for those jobs.
- The PR gets stuck in a "waiting" state (often shown as "Expected — Waiting for a status to be reported" in the PR checks or branch protection UI). This blocks merges indefinitely, even though the changes are "safe" (only docs).

- This doesn't happen on every PR—only when all changes match paths-ignore.

Because of above limitation i replaced `path-ignore ` with `dorny/paths-filter@v3`

tiagolobocastro: Ah that's right indeed I remember this issue now, I'll revert and make similar change to zfs as well.

niladrih (approved):

Issue #461, opened 2026-03-18
NodeUnpublishVolume reports success after EBUSY, leaving stale mount and breaking next NodePublishVolume


Liamlu28:
## Bug report

We hit a stale mount issue with `openebs/lvm-driver:1.8.0` (`local.csi.openebs.io`) on a Kubernetes production cluster.

The main problem is:

- `NodeUnpublishVolume` hits `device or resource busy`
- but the driver still logs `umount done` / `has been unmounted`
- the old mount actually remains on the host
- the next `NodePublishVolume` then fails with:
  `verifyMount: device already mounted at ...`

This leaves StatefulSet rollouts stuck until we manually clean the stale mount from the host mount namespace.

## Environment

- Driver: `docker.io/openebs/lvm-driver:1.8.0`
- CSI driver: `local.csi.openebs.io`
- StorageClass: `localssd`
- Access mode: `RWOP`
- Node-local LVM volumes
- Kubernetes: production cluster v1.33.7
- Workload type: StatefulSet rolling update

## What happened

During a rolling update, old pods were deleted, and replacement pods were scheduled on the same node.

The driver received `NodeUnpublishVolume` for the old pod path, but unmount cleanup was not completed correctly. The driver then proceeded as if unpublish had succeeded.

Immediately after that, `NodePublishVolume` for the replacement pod failed because the device was still mounted at the old pod path.

This blocked the new pod from mounting the same PVC.

### NodeUnpublishVolume
```text
GRPC call: /csi.v1.Node/NodeUnpublishVolume requests {"target_path":"/var/lib/kubelet/pods/9c76d145-5a55-45ac-b13c-b2323e62c048/volumes/kubernetes.io~csi/pvc-96e56831-c0c1-4c70-b06c-709dd5806565/mount","volume_id":"pvc-96e56831-c0c1-4c70-b06c-709dd5806565"}

E... mount.go:136] lvm: failed to remove mount path vol pvc-96e56831-c0c1-4c70-b06c-709dd5806565 err : remove /var/lib/kubelet/pods/9c76d145-5a55-45ac-b13c-b2323e62c048/volumes/kubernetes.io~csi/pvc-96e56831-c0c1-4c70-b06c-709dd5806565/mount: device or resource busy

[...]

PR #469, opened 2026-04-20
fix: propagate unmount errors in NodeUnpublishVolume (#461)

kotyara85:
Replace hand-rolled unmount + os.Remove logic in UmountVolume with mount.CleanupMountPoint from k8s.io/mount-utils.

Previously, UmountVolume swallowed os.Remove errors after a successful unmount (logging but not returning them), causing NodeUnpublishVolume to report success even when the mount point directory could not be cleaned up. This left stale mount points that broke subsequent NodePublishVolume calls, manifesting as pods stuck in ContainerCreating.

CleanupMountPoint properly handles all edge cases (corrupted mounts, EBUSY, path removal) and propagates every error to the caller so kubelet can retry.

### DIFF
--- a/pkg/lvm/mount.go
+++ b/pkg/lvm/mount.go
@@ -17,7 +17,6 @@ package lvm
 
 import (
 	"errors"
-	"fmt"
 	"math"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


hxydxn:
We hit this exact bug in production with `lvm-driver:1.8.1`. The sequence matches #461:

1. Pod killed due to liveness probe failure
2. `NodeUnpublishVolume` called — `mounter.Unmount(targetPath)` succeeds but `os.Remove(targetPath)` fails with `EBUSY` (device or resource busy)
3. Driver logs `umount done` and returns success despite the error
4. New pod scheduled to the same node — `NodePublishVolume` fails with `verifyMount: device already mounted at [...]`
5. Pod stuck in `ContainerCreating` indefinitely

In our case, the nodes have `/var/lib/kubelet` bind-mounted from a larger data partition (`/mnt/local/kubelet`), so the device appeared at two mount paths. The `os.Remove` failed because only one layer was unmounted.

We had to manually SSH to each affected node and `umount` the stale paths to recover.

+1 for merging this.

tiagolobocastro (approved):

kotyara85:
@tiagolobocastro Thanks for approving, this did fix the described issue in prod for us. Do you know when this can be merged and released? Ty!

rohan2794 (approved):

abhilashshetty04:
@mergify queue

tiagolobocastro:
> @tiagolobocastro Thanks for approving, this did fix the described issue in prod for us. Do you know when this can be merged and released? Ty!

We were planning on releasing soonish, maybe in the coming weeks
If you need a patch release faster, we could cherry-pick this into release/1.8 branch?

kotyara85:
> > @tiagolobocastro Thanks for approving, this did fix the described issue in prod for us. Do you know when this can be merged and released? Ty!
> 
> We were planning on releasing soonish, maybe in the coming weeks If you need a patch release faster, we could cherry-pick this into release/1.8 branch?

Yeah, that's what we did :). ty!

PR #470, opened 2026-04-20
fix(controller): fallback to VG capacity if thin pool is missing

rohan2794:
When `thinProvision` is enabled and the StorageClass uses `volumeBindingMode: WaitForFirstConsumer`, the `GetCapacity` call returns `0` if the expected thin pool does not already exist on the node. However, because the scheduler sees `0` capacity during the scheduling cycle, it never selects a node, and `CreateVolume` is never triggered. This creates a deadlock for new thin-provisioned volumes.

**What this PR does?**:
Modified `GetCapacity` to report the Volume Group's free capacity as a fallback if the thin pool is not found. This allows the scheduler to bind the volume to a viable node, subsequently triggering the `CreateVolume` call where the thin pool will be automatically provisioned.

No

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -971,22 +971,12 @@ func (cs *controller) GetCapacity(
 			vgThinPoolName := fmt.Sprintf("%s_thinpool", vg.Name)
 			freeCapacity := vg.Free.Value()
 			if params.ThinProvision == lvm.YES {
-				foundPool := false
 				for _, pool := range vg.ThinPools {
 					if pool.Name == vgThinPoolName {
 						freeCapacity += pool.Free.Value()
-						foundPool = true
 						break
 					}
 				}
-				// If thin pool isn't found, report VG free capacity
-				// because the driver will create the pool automatically.
-				if !foundPool {
-					klog.V(4).Infof("Thin pool %s not found on node %s, falling back to VG free capacity: %d",
-						vgThinPoolName, nodeName, vg.Free.Value())
-				}
-			} else {
-				freeCapacity = vg.Free.Value()
 			}
 			if availableCapacity < freeCapacity {
 				availableCapacity = freeCapacity
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -968,14 +968,22 @@ func (cs *controller) GetCapacity(
 			if !params.VgPattern.MatchString(vg.Name) {
 				continue
 			}
-
-			var freeCapacity int64
+			vgThinPoolName := fmt.Sprintf("%s_thinpool", vg.Name)
+			freeCapacity := vg.Free.Value()
 			if params.ThinProvision == lvm.YES {
+				foundPool := false
 				for _, pool := range vg.ThinPools {
-					if pool.Name == vg.Name+"_thinpool" {
-						freeCapacity = pool.Free.Value()
+					if pool.Name == vgThinPoolName {
+						freeCapacity += pool.Free.Value()
+						foundPool = true
 						break
 					}
+				}
+				// If thin pool isn't found, report VG free capacity
+				// because the driver will create the pool automatically.
+				if !foundPool {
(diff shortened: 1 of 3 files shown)


pkg/driver/controller.go
```diff
@@ -971,12 +971,21 @@ func (cs *controller) GetCapacity(
 
 			var freeCapacity int64
 			if params.ThinProvision == lvm.YES {
+				foundPool := false
 				for _, pool := range vg.ThinPools {
 					if pool.Name == vg.Name+"_thinpool" {
 						freeCapacity = pool.Free.Value()
```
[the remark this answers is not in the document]
rohan2794: [fixed](https://github.com/openebs/lvm-localpv/compare/5bfd5df3c35e35edd92870cece9e4fa3c4c000b9..31cdf39dc50708766176b0941a31a662c3d23cc6)

pkg/driver/controller.go
```diff
@@ -971,12 +971,21 @@ func (cs *controller) GetCapacity(
...
+						foundPool = true
 						break
 					}
 				}
+				// If thin pool isn't found, report VG free capacity
+				// because the driver will create the pool automatically.
+				if !foundPool {
+					klog.Infof("Thin pool %s_thinpool not found on node %s, falling back to VG free capacity: %d",
```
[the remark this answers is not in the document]
rohan2794: [fixed](https://github.com/openebs/lvm-localpv/compare/5bfd5df3c35e35edd92870cece9e4fa3c4c000b9..31cdf39dc50708766176b0941a31a662c3d23cc6)

pkg/driver/controller.go
abhilashshetty04: Should we add vg free space and thinpool free space if thinpool exist? Because vg free space can be claimed by thinpool using auto extend.

https://github.com/openebs/lvm-localpv/blob/develop/pkg/driver/schd_helper.go#L137

rohan2794: [considered vg free capacity ](https://github.com/openebs/lvm-localpv/compare/31cdf39dc50708766176b0941a31a662c3d23cc6..a37ffacdf5efd14401bc014f6bd8fed428edba19)

pkg/driver/controller.go
```diff
@@ -968,15 +968,23 @@ func (cs *controller) GetCapacity(
 			if !params.VgPattern.MatchString(vg.Name) {
 				continue
 			}
-
-			var freeCapacity int64
+			vgThinPoolName := fmt.Sprintf("%s_thinpool", vg.Name)
+			freeCapacity := vg.Free.Value()
```
tiagolobocastro, line 972: this feels a bit confusing now as we have the else below setting the same thing
Probably we don't need the else anymore?

pkg/driver/controller.go
```diff
@@ -968,15 +968,23 @@ func (cs *controller) GetCapacity(
...
-						freeCapacity = pool.Free.Value()
+					if pool.Name == vgThinPoolName {
+						freeCapacity += pool.Free.Value()
+						foundPool = true
 						break
 					}
 				}
+				// If thin pool isn't found, report VG free capacity
+				// because the driver will create the pool automatically.
+				if !foundPool {
```
abhilashshetty04: Do we need this need to be logged? GetCapacity is invoked periodically, this might flood the log.

rohan2794: removed it

pkg/driver/controller.go
rohan2794: [fixed](https://github.com/openebs/lvm-localpv/pull/470/changes/eb3cd8c902e962335612f25294848e1575fa64f8)

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

tiagolobocastro:
@mergify queue

PR #493, opened 2026-07-23
fail CreateVolume api if both vgPattern and volGroup is passed

abhilashshetty04:
volgroup and vgpattern are mutually exclusive: volgroup pins the volume to a single volume group, whereas vgpattern selects among many. 
Supplying both is ambiguous, so reject the request.

### DIFF
--- a/docs/storageclasses.md
+++ b/docs/storageclasses.md
@@ -221,8 +221,6 @@ LocalPV-LVM storageclass supports various parameters for different use cases. Fo
     vgpattern: "lvmvg.*"     ## vgpattern specifies pattern of lvm volume group name
   ```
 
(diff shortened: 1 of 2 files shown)


rohan2794 (approved):

tiagolobocastro (commented):
Isn't this a breaking change if somewhere is already using SC with this?

abhilashshetty04:
> Isn't this a breaking change if somewhere is already using SC with this?

Yes it is. Event on Pending PVC will show the error we return. We can highlight this change in the release notes as well. 

Will it need version bump?

abhilashshetty04:
This will be merged when we are ready to release v2.0.

PR #506, opened 2026-09-01
add default format options to storage class

abhilashshetty04:
Adds a node-agent flag `--default-format-options=<fstype>=<options>` (helm: `lvmNode.defaultFormatOptions`)
that supplies extra `mkfs` options per filesystem when a volume's StorageClass does not set
`formatOptions`. A StorageClass value *replaces* the default for its filesystem — the two are
never merged.

`mkfs.xfs` 6.5+ (what the driver image now ships) enables `nrext64` on new filesystems, and only
kernel 5.19+ can mount one. On RHEL 8/9, Ubuntu 20.04/22.04 or SLES 15 nodes, an xfs volume
formats fine and then fails to mount with:

    XFS (dm-0): Superblock has unknown incompatible features (0x20) enabled.

Users can already work around this per StorageClass via `formatOptions`, but that requires every
user to know about it. This makes the safe value the default, cluster-wide, and lets clusters on
5.19+ opt out with `--set lvmNode.defaultFormatOptions.xfs=""`.

Defaults only apply at format time (first use), so changing the value does not affect existing
volumes.

Referred PR: https://github.com/openebs/zfs-localpv/pull/760


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

abhilashshetty04:
@mergify backport release/1.10

